PR #60, opened 2020-05-29
Post Block Component

urchykli:


urchykli:
@jnschrag I'm not sure how to make this change without completely changing everything about the grid layout, so can we chat about this tomorrow?
> * For the post date & author HTML, I would recommend a structure like this. It will also simplify the CSS a little bit.
> 
> ```
> <div class="post-meta post-block__date">
>   <span class="post-meta__label">Published</span>
>   {{ formatDate }}
> </div>
> <div class="post-meta post-block__author">
>   <span class="post-meta__label">Written by</span>
>   {{ getAuthor}}
> </div>
> ```

components/global/PostBlock.vue
```diff
@@ -1,12 +1,26 @@
...
-  <article class="post-component post-component--wide">
-    <header>
-      <h2>
-        <nuxt-link :to="nestedSlug">{{ data.title }}</nuxt-link>
-      </h2>
-      {{ data.postDate }}
-    </header>
-    {{ data.excerpt }}
+  <article class="post-component post-component--wide post-block">
+    <div class="post-block__category">
```
jnschrag, line 3: Can you update this so it only renders if there is a category? Right now, the `<div>` renders no matter what, but there's no need for an empty div if there's no category assigned to it.

components/global/PostBlock.vue
```diff
@@ -1,12 +1,26 @@
...
+    <img class="post-block__image" src="http://placekitten.com/200/138" />
+    <h2 class="post-block__title">
+      <nuxt-link :to="nestedSlug">{{ data.title }}</nuxt-link>
+    </h2>
+    <div class="post-meta">
+      <div class="post-meta__block post-block__date">
+        <span class="post-meta__label">Published</span>
+        {{ formatDate }}
+      </div>
+      <div class="post-meta__block post-block__author">
```
jnschrag, line 16: Same thing for the authors. If no author is provided, there shouldn't be a div here.

assets/css/components/post-block.scss
```diff
@@ -0,0 +1,136 @@
...
+    }
+  }
+
+  &__title {
+    grid-column: 1 / span 1;
+    grid-row: 2 / span 1;
+    margin-top: rem(4px);
+    color: $color-white-190;
+    @extend %font-ui-heading-post-block-title;
+    font-weight: 500;
```
jnschrag, line 35: If the font weight is supposed to be 500, can you update it in the `_placeholders.scss` file instead of adding it separately here?

assets/css/components/post-block.scss
```diff
@@ -0,0 +1,136 @@
...
+    }
+  }
+
+  &__excerpt {
+    grid-column: 1 / span 1;
+    grid-row: 4 / span 1;
+    margin: rem(17px) 0 rem(32px) 0;
+    color: $color-white-180;
+    @extend %font-ui-text-lg-oxygen;
+    line-height: 150%;
```
jnschrag, line 57:
```suggestion
    line-height: 1.5;
```

Best practice is to use unitless line-heights so they scale.

assets/css/components/post-block.scss
```diff
@@ -0,0 +1,136 @@
...
+    @include breakpoint($break: 'small', $dir: 'max-width') {
+      margin-top: rem(16px);
+      margin-left: 0;
+    }
+  }
+
+  &__excerpt {
+    grid-column: 1 / span 1;
+    grid-row: 4 / span 1;
+    margin: rem(17px) 0 rem(32px) 0;
```
jnschrag, line 54:
```suggestion
    margin: rem(16px) 0 rem(32px) 0;
```

Tucker always uses a grid of 8, so if you get a weird number like 17 or 22, round it to the closest multiple of 8.

assets/css/components/post-block.scss
```diff
@@ -0,0 +1,136 @@
+.post-block {
+  position: relative;
+  display: grid;
+  grid-template-columns: 70% 30%;
```
jnschrag, line 4:
```suggestion
  grid-template-columns: 2fr 1fr;
```

Where possible, when using CSS grid use the `fr` unit. It takes into account the grid gap when assigning space.

assets/css/components/post-block.scss
```diff
@@ -0,0 +1,136 @@
...
+      }
+    }
+  }
+}
+
+.post-meta {
+  display: flex;
+  grid-column: 1 / span 1;
+  grid-row: 3 / span 1;
+  margin: rem(9px) 0 0 0;
```
jnschrag, line 120:
```suggestion
  margin: rem(8px) 0 0 0;
```

assets/css/components/post-block.scss
```diff
@@ -0,0 +1,105 @@
...
+
+    @include breakpoint('medium') {
+      grid-column: 1 / span 1;
+      margin: 0;
+    }
+
+    .visualizer {
+      @extend %font-ui-heading-section-md;
+      padding-bottom: rem(2px);
+      color: $color-white-190;
```
jnschrag, line 50: The color is common in both places it's used, so you don't need to specify it in the `.visualizer` class.

components/global/PostBlock.vue
```diff
@@ -17,22 +32,45 @@ export default {
...
+          featureImageCaption: ''
         }
       }
     }
   },
   computed: {
     nestedSlug() {
       return `analysis/${this.data.slug}`
+    },
+    getImage() {
```
jnschrag, line 51: This property can be removed since we no longer have featured images.

Issue #17, opened 2020-02-13
Layout: Post


jnschrag:
Create a page for the individual analysis posts.

Page should include:
- [x] Category
- [x] Post Title
- [x] Excerpt
- [x] Satellite Table
- [x] Authors List
- [x] Published Date
- [x] Last Updated Date
- [x] Feature Image
- [x] Content
- [x] Further Reading
- [x] Authors Block
- [x] Tags
- [x] Related Posts
- [x] Site Footer

jnschrag:
This probably should have been split into multiple issues, but...oh well. :D

@urchykli I'd like you to start with the page structure & the page header. For the structure, I mean the HTML required for all the different components with placeholder text indicating what goes where. (eg. `<div class="related-posts">Related Posts</div>`)

<img width="736" alt="Screen Shot 2020-06-08 at 2 22 23 PM" src="https://user-images.githubusercontent.com/1896496/84066205-9008bd80-a993-11ea-858e-336bf4c42e49.png">

urchykli:
Questions
- Will updated dates and content be part of the post content or will it be a separate piece of data?
- What's the path to access author bios using the author id?
- What's the path to access related post content using the related post id?

Data Needed
- Satellite
     -user (gov't, commercial, etc)
     -URL
     -Object name
- Updated content (depending on the answer to the question above)
- Tags
     -Name
     -URL
- Further reading
     -Title
     -Source
     -URL
-Related posts (2 per post)
     -Featured image
     -Title
     -Author
     -Date
     -URL

jnschrag:
@shortjared For things like the featured images, do we have the URLs for what those images will be yet?

PR #66, opened 2020-06-24
adds additional examples of accessing data

shortjared:
This should address quite a few of the things in issue #17, the big one I still need to get you is featured images.


urchykli:
Thank you @shortjared!

PR #67, opened 2020-06-25
Single Post Component

urchykli:
Questions/comments:
- Margins/padding on mobile are pushing the content all to the left side.
- How do I ignore the padding within the content for images and blockquotes?  
- When we talked before, we discussed using `.entry-content` to style the elements within the content, but there was already some styling attached to `.post__content`.  I moved the styles from `.post__content` to `.entry-content`, but wanted to make sure that was correct before deleting the `.post__content` styles.
- I haven't styled the updated section because I wasn't sure if it would appear in the content or as a standalone piece of data.

Data needed within the content:
- Image
- Image caption
- Table
- Video
- Footnotes
- Updated content and date

Data needed from my comment on the issue:
- Satellite User
- Satellite URL
- Satellite/object name
- Featured Image


jnschrag:
Commenting to add that if you want to go over the html structure I mentioned above, let's schedule a time next week to review it together. I remembered this morning that this is one of your first times laying out a page in full like this, so I want to make sure you have the chance to ask questions if needed!

assets/css/pages/post.scss
```diff
@@ -1,15 +1,353 @@
...
+  }
+
+  &__sat-table {
+    @extend %font-ui-text-base-oxygen;
+    width: 100%;
+    margin-top: rem(24px);
+    border-collapse: collapse;
+
+    thead {
+      background-color: #181b1f;
```
jnschrag, line 104:
```suggestion
      background-color: $color-slate-500;
```

assets/css/pages/post.scss
```diff
@@ -1,15 +1,353 @@
...
+      @extend %font-post-text-footnote-list;
+      padding: rem(64px) 0 rem(56px) 0;
+      color: $color-black-160;
+      border-bottom: rem(1px) solid $color-black-120;
+    }
+
+    &-header {
+      @extend %font-ui-heading-section-lg;
+      margin: rem(32px) 0 rem(16px) 0;
+      color: #0c1013;
```
jnschrag, line 181:
```suggestion
      color: $color-slate-800;
```

assets/css/pages/post.scss
```diff
@@ -1,15 +1,353 @@
...
+    }
+
+    &-link {
+      z-index: 2;
+      display: table-cell;
+      padding: rem(24px) rem(32px);
+      vertical-align: middle;
+
+      &:hover .post__related-title {
+        padding-bottom: rem(12px);
```
jnschrag, line 332: Typically, it's not a good idea to adjust margin or padding values on hover because it forces the browser to redraw the layout. I would recommend using applying `transform: translate(0, rem(-4px))` on the title instead, which will move the title up a bit (creating more space between it and the text below it) but won't trigger a redraw. Remember to add a transition to it as well so it isn't a jarring movement.

pages/analysis/_slug.vue
```diff
@@ -1,19 +1,126 @@
...
+              <a href="#">
+                {{ t.tag.name }}
+              </a>
+            </li>
+          </ul>
+        </div>
+      </footer>
+    </div>
+    <section>
+      <div class="post__related-wrapper">
```
jnschrag, line 92: I'm not sure why there's both a `<section>` element and a `<div>` here. I think you could just use the `<section>` element.

PR #70, opened 2020-07-31
Feat/site nav

urchykli:
I *think* I did this right, but please let me know if you need me to clean up anything.  Thanks!


components/global/Navigation.vue
```diff
@@ -0,0 +1,67 @@
+<template>
+  <div class="nav" :class="parentClass">
+    <p class="is-footer">Navigate</p>
+    <ol :class="parentClass + '__list'">
```
jnschrag, line 4: Navigation lists are usually unordered lists instead of ordered.

components/global/Navigation.vue
```diff
@@ -0,0 +1,67 @@
+<template>
+  <div class="nav" :class="parentClass">
+    <p class="is-footer">Navigate</p>
```
jnschrag, line 3: Can we use a more appropriate class name for this? Something like `nav__title` so it's more descriptive of what it does?

[Take a look](https://vuejs.org/v2/guide/conditional.html) at using a `v-if` statement to selectively render that HTML. Hiding it with CSS works, but if we can just not render the HTML, that's even better.

components/global/Navigation.vue
```diff
@@ -0,0 +1,67 @@
...
+          title: 'Glossary',
+          url: '/glossary',
+          id: 'glossary'
+        }
+      ]
+    }
+  },
+  computed: {
+    parentClass() {
+      return `nav-${this.name}`
```
jnschrag, line 59: For proper BEM naming, this is technically a modifier, so it should have two dashes, like this `nav--${this.name}`.

assets/css/components/navigation.scss
```diff
@@ -0,0 +1,43 @@
...
+    }
+
+    &__item:not(:last-child) {
+      margin-bottom: rem(8px);
+    }
+  }
+
+  &-header,
+  &-footer {
+    &__item {
```
jnschrag, line 37: Does this need to be nested inside `&-header/&-footer` or can it just be `&__item` (which compiles to `.nav__item`)?

components/global/Navigation.vue
urchykli: Fixed! (the issue said `ol` so I went with it)

assets/css/components/navigation.scss
```diff
@@ -0,0 +1,36 @@
...
+  color: $color-white-170;
+  text-transform: capitalize;
+
+  &__item {
+    &:hover {
+      color: $color-white-180;
+    }
+  }
+
+  &-header {
```
jnschrag, line 12:
```suggestion
  &--header {
```

PR #72, opened 2020-08-14
Attempted to style footer according to Figma

Jonathan-J-Choi:


components/global/Footer.vue
```diff
@@ -1,23 +1,79 @@
...
+        },
+        {
+          title: 'University of Texas at Austin',
+          url: 'https://www.utexas.edu/',
+          id: 'universtiy-of-texas-at-austin'
+        }
+      ]
+    }
+  },
+  computed: {
```
jnschrag, line 69: You can remove this computed property. It was used in the navigation component, but really isn't necessary here.

components/global/Footer.vue
```diff
@@ -1,23 +1,79 @@
...
     Navigation
+  },
+
+  data() {
+    return {
+      items: [
+        {
+          title: 'Center for Strategic and International Studies',
+          url: 'https://www.csis.org',
+          id: 'center-for-strategic-and-international-studies'
```
jnschrag, line 54: For the ids, you can use the acronyms instead of the full names. Eg. `csis`, `swf`, and `uta`.

components/global/Footer.vue
```diff
@@ -1,23 +1,79 @@
...
+    <p class="footer__description">
+      Lorem ipsum dolor sit amet consectetur adipisicing elit. Laborum dolores
+      modi adipisci inventore, nihil cumque. Cumque, sit ipsam? Est aperiam
+      doloremque esse nihil? Asperiores provident ullam incidunt qui nostrum
+      nam!
+    </p>
     <Social />
+    <div class="wrapper">
+      <Navigation name="footer" />
+      <div class="footer__learnMoreAbout" :class="parentClass">
```
jnschrag, line 13: We've been calling this section "Learn More About", but one of the dangers of naming things based off the written text is if we decide down the line to change the heading to something other than "Learn More About", these classes no longer make sense. My recommendation is to think about about functionality when thinking about names. In this case, I think `footer__organizations` fits better because that's what this section is - a list of organizations involved.

You can remove the `:class="parentclass"` bit from here too.

components/global/Footer.vue
```diff
@@ -1,23 +1,79 @@
...
+        <h2 class="footer__learnMoreAboutHeader">
```
jnschrag, line 14: This is another example where we have to be cautious about getting too specific with our class names. `footer__heading` or `footer__subtitle` would be more appropriate here. The other thing to look at is if it has any similarities with other elements in the same component. For example, the heading for the navigation & the organizations list are stylistically exactly the same, so it would make sense that they use the same CSS class. Otherwise you end up duplicating the styles unnecessarily.

To do this, you'll want to change the Navigation component so it uses `<h2 class="footer__subtitle">` too. The styles should go in `footer.scss` because it's the parent component.

components/global/Footer.vue
```diff
@@ -1,23 +1,79 @@
...
+          Learn More About
+        </h2>
+        <ul :class="parentClass + '__list'">
+          <li
+            v-for="item in items"
+            :key="item.id"
+            :class="[parentClass + '__' + item.id, parentClass + '__item']"
+            class="footer__learnMoreAbout__item"
```
jnschrag, line 22: To match BEM naming, you'd want to make this `footer__organizations-item`.

components/global/Footer.vue
```diff
@@ -1,23 +1,79 @@
 <template>
   <footer class="footer">
-    <Navigation name="footer" />
+    <SiteLogo />
+    <p class="footer__description">
```
jnschrag, line 4: You actually don't need to have a CSS class on this paragraph (but if you did, it's common to use `desc` as a shorthand for description eg. `footer__desc`). Because it's the only one used, you can use a more generic selector of `.footer p` to style it.

PR #73, opened 2020-09-01
Button styling, working on the switching between a and button tags.

### LINKED ISSUE #27: Component: Button
Create a button component that can be used throughout the site. See the SC War Game for an [example](https://github.com/CSIS-iLab/sc-trade-war-game/blob/master/components/Button.vue).

- [ ] Switch between `a` and `button` tags. See [this article](https://www.codegram.com/blog/create-a-reusable-button-with-vue-dynamic-components/) for reference
- [ ] Set default `btn` class
- [ ] Take `onClick` prop
- [ ] Take `href` prop

## Button Types
The `btn` class should hold styles that are common for all buttons. Modifier classes should be used for all variants of a button & only change the properties necessary to modify the base class. Multiple modifiers can be used on a single button. For example, an outlined round light button would have all of these classes applied, `btn btn--outlined btn--round btn--light`.
- [ ] `.btn`: Text only, large text
- [ ] `.btn--sm`: Small text
- [ ] `.btn--contained`: Has background
- [ ] `.btn--round`: Has rounded corners
- [ ] `.btn--outlined`: Has border
- [ ] `.btn--light`: Light background/text
- [ ] `.btn--icon`: Only contains an icon

## Button States
For each button type above, these states need to be accounted for:
- [ ] Base
- [ ] Hover
- [ ] Focus (note that the focus state should be handled either with the `outline` or `box-shadow` properties, not `border`)
- [ ] Disabled

Jonathan-J-Choi:
Addressed styling changes as instructed.

- resolved issues as specified by Jacque regarding: padding of buttons

- removed commented code from button.vue

- removed pointer-events:none

- added transition to .btn

Added comment on the css compiling issue for btn--small. closes #27

### DIFF
--- a/assets/css/components/button.scss
+++ b/assets/css/components/button.scss
@@ -1,6 +1,11 @@
 .btn {
   @extend %font-ui-text-base-good;
+  padding: 0 10px;
   color: $color-white-160;
+  background-color: transparent;
+  border-style: none;
+  border-radius: 0;
+  opacity: 0.7;
 
   &:hover {
     color: $color-white-180;
@@ -8,34 +13,29 @@
 
   &:focus {
     box-sizing: border-box;
-    box-shadow: 1px $color-focus;
+    outline: none;
+    box-shadow: 0 0 0 2px $color-focus;
   }
 
   &:disabled {
     color: $color-white-160;
+    cursor: not-allowed;
+    opacity: 0.4;
+    pointer-events: none;
   }
 
   &--sm {
-    @extend %font-ui-text-base-good;
+    @extend %font-ui-text-sm-good;
   }
 
   &--contained {
     color: $color-white-170;
     background-color: $color-slate-200;
-    border-radius: 3px;
+    border-radius: 50px;
 
     &:hover {
       color: $color-white-190;
       background-color: $color-slate-100;
-    }
-
-    &:focus {
-      box-sizing: border-box;
-      box-shadow: 1px $color-focus;
-    }
-  
-    &:disabled {
-      color: $color-white-160;
     }
   }
 
@@ -47,15 +47,6 @@
     &:hover {
       color: $color-white-190;
       background-color: $color-slate-700;
-    }
-
-    &:focus {
-      box-sizing: border-box;
-      box-shadow: 1px $color-focus;
-    }
-  
-    &:disabled {
-      color: $color-white-160;
     }
   }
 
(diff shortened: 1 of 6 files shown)


assets/css/components/button.scss
```diff
@@ -0,0 +1,114 @@
...
+      color: $color-slate-100;
+    }
+
+    &:focus {
+      box-sizing: border-box;
+      box-shadow: 1px $color-focus;
+    }
+  }
+
+  .btn--icon {
```
jnschrag, line 97:
```suggestion
  &--icon {
```

components/global/Button.vue
```diff
@@ -0,0 +1,53 @@
+<template>
+  <content :is="type" :href="href" class="button btn" @click="onClick">
```
jnschrag, line 2:
```suggestion
  <content :is="type" :href="href" class="btn" @click="onClick">
```

Removing the unneeded `button` class.

assets/css/components/button.scss
```diff
@@ -0,0 +1,114 @@
...
+    box-sizing: border-box;
+    box-shadow: 1px $color-focus;
+  }
+
+  &:disabled {
+    color: $color-white-160;
+  }
+
+  &--sm {
+    @extend %font-ui-text-base-good;
```
jnschrag, line 19:
```suggestion
    @extend %font-ui-text-base-sm-good;
```
This should be different from the default `.btn`

assets/css/components/button.scss
```diff
@@ -0,0 +1,81 @@
+.btn {
+  @extend %font-ui-text-base-good;
+  padding: 0 10px;
```
jnschrag, line 3:
```suggestion
  padding: 4px 12px;
```

Sorry, changed my mind after seeing it implemented!

components/global/Button.vue
```diff
@@ -0,0 +1,53 @@
...
+      if (this.href) {
+        return 'a'
+      } else {
+        return 'button'
+      }
+    }
+  }
+}
+</script>
+<!--<template>
```
jnschrag, line 34: Can you remove this commented out code, please?

assets/css/components/button.scss
```diff
@@ -0,0 +1,81 @@
...
+    box-sizing: border-box;
+    outline: none;
+    box-shadow: 0 0 0 2px $color-focus;
+  }
+
+  &:disabled {
+    color: $color-white-160;
+    cursor: not-allowed;
+    opacity: 0.4;
+    pointer-events: none;
```
jnschrag, line 24:
```suggestion
```

TIL that if you have `pointer-events: none` set, it won't show the cursor at all. I didn't realize that!

assets/css/components/button.scss
```diff
@@ -0,0 +1,81 @@
+.btn {
+  @extend %font-ui-text-base-good;
+  padding: 0 10px;
+  color: $color-white-160;
+  background-color: transparent;
+  border-style: none;
+  border-radius: 0;
+  opacity: 0.7;
```
jnschrag, line 8:
```suggestion
  opacity: 0.7;
  transition: all 0.3s ease-in-out;
```

The last thing this needs is a transition effect so that when the state of the button changes, it will smoothly transition the text, background, & border colors. Generally speaking, it's best to specify which properties need to be transitioned, but in the case of these buttons (when it can vary), it's acceptable to set it to `all`.

jnschrag (changes requested):
This is looking really good! Really nice job in implementing this. I'm going to give you a call to show you one additional thing that I can explain better over video than in a comment, but this is great!

jnschrag (approved):
Nice work!

PR #74, opened 2020-09-04
laying out framework for review

### LINKED ISSUE #32: Component: Text Input
Create reusable text input component. It would be worth looking at how the Carbon Design System implements this. [Storybook Example](http://vue.carbondesignsystem.com/?path=/story/components-cvtextinput--default) & [GitHub Code](https://github.com/carbon-design-system/carbon-components-vue/blob/master/packages/core/src/components/cv-text-input/cv-text-input.vue)

- [ ] States
  - [ ] Base
  - [ ] Disabled
  - [ ] Hover
  - [ ] Focus
  - [ ] Error

Jonathan-J-Choi:
Implement styles for input boxes & labels. Closes #32.

### DIFF
--- a/assets/css/components/forms.scss
+++ b/assets/css/components/forms.scss
@@ -25,7 +25,7 @@
     opacity: 0.4;
   }
 
-  .input--error {
+  &--error {
     box-sizing: border-box;
     outline: none;
     box-shadow: 0 0 0 2px $color-orangedark-100;
--- a/assets/css/components/forms.scss
+++ b/assets/css/components/forms.scss
@@ -1,9 +1,10 @@
-.textInput {
+.input {
   @extend %font-ui-text-base-oxygen;
   color: $color-white-150;
   background-color: $color-slate-800;
   border-style: none;
   border-radius: 3px;
+  transition: all 0.3s ease-in-out;
 
   &:hover {
     background-color: $color-slate-900;
@@ -22,16 +23,33 @@
   &:disabled {
     cursor: not-allowed;
     opacity: 0.4;
-    pointer-events: none;
   }
 
-  &:error {
+  .input--error {
     box-sizing: border-box;
     outline: none;
     box-shadow: 0 0 0 2px $color-orangedark-100;
   }
 }
 
-.test {
-  color: $color-white-150;
+.label {
+  @extend %font-ui-text-base-oxygen;
+  color: $color-white-160;
+  border-style: none;
+  border-radius: 3px;
+  transition: all 0.3s ease-in-out;
+
+  &:hover {
+    color: $color-white-180;
+  }
+
+  &:focus {
+    box-sizing: border-box;
+    color: $color-white-140;
+    background-color: $color-slate-800;
+    border-radius: 50px;
+    outline: none;
+    box-shadow: 0 0 0 2px $color-focus;
+  }
 }
+
--- a/assets/css/components/text-input.scss
+++ b/assets/css/components/text-input.scss
@@ -1,0 +1,37 @@
+.textInput {
+  @extend %font-ui-text-base-oxygen;
+  color: $color-white-150;
+  background-color: $color-slate-800;
+  border-style: none;
+  border-radius: 3px;
+
+  &:hover {
+    background-color: $color-slate-900;
+  }
+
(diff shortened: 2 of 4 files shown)


assets/css/components/text-input.scss
```diff
@@ -0,0 +1,37 @@
...
+    color: $color-white-180;
+  }
+
+  &:disabled {
+    cursor: not-allowed;
+    opacity: 0.4;
+    pointer-events: none;
+  }
+
+  &:error {
```
jnschrag, line 28: Unfortunately, `error` isn't a valid pseudo-class, so you'll ned to apply these styles with another class. My recommendation is to do something like `.input--error` instead.

assets/css/components/text-input.scss
```diff
@@ -0,0 +1,37 @@
+.textInput {
```
jnschrag, line 1: Can you make this `.input` instead? The reason why is we're going to want to use this styles on a variety of inputs, not all of which are your standard `input[type="text"]` fields and we'll want to use the classes on those as well.

components/global/TextInput.vue
```diff
@@ -0,0 +1,20 @@
+<template>
+  <div>
+    <input v-model="message" class="textInput" placeholder="Placeholder Text" />
+    <p class="test">Message: {{ message }}</p>
```
jnschrag, line 4: In order for this to be semantic, it should actually be a `<label>` tag and should be connected to the input via an ID property.

Note that inputs should typically have an id & a name field, so we should add those as props. I would set it so if "name" is not defined, it defaults to the ID.

jnschrag (changes requested):
This is a really solid start! Thank you for working on it. :D

assets/css/components/forms.scss
```diff
@@ -0,0 +1,55 @@
...
+  &:active {
+    color: $color-white-180;
+  }
+
+  &:disabled {
+    cursor: not-allowed;
+    opacity: 0.4;
+  }
+
+  .input--error {
```
jnschrag, line 28:
```suggestion
  &--error {
```
You can use the `&` here instead of typing out `.input` again.

jnschrag (changes requested):
Thanks so much for working through this, Jonathan! I think we have a solid foundation to work off of, but there's a few things that are missing here. Note that I resolved a merge conflict & added some properties to the `<textInput />` example so I could see how it all worked together.

## Styling
- [x] Please add padding to the input field per the mockups
- [ ] The `color` property on `.input` should be `$color-white-180`. To get the gray color of the placeholder text, you'll need to use the [placeholder](https://developer.mozilla.org/en-US/docs/Web/CSS/::placeholder) pseudo element (eg. `&::placeholder { color: $color-white-150; }`
- [ ] Labels don't have a hover or focus state themselves (when you click on them, it focuses on their accompanying input). The exception to this is if the `<label>` text contains a button, in which case the button needs a focus state.
  - [ ] Likewise, the label doesn't need a border, border-radius, or transition properties defined.

## HTML
- [ ] The `<label>` tag is missing `class="label"`.
- [ ] I would investigate using the `<slot>` tag, specifically for the labels. The reason being is that some labels require a button to be a child of the label. Happy to talk through this with you next week if you want!
- [ ] The input is missing the clear button. Fun fact, `input type="search"` includes that button by default, but since we're creating a text input and not a search, we need to make it ourselves (womp). Here's what I would recommend for handling that:
  - Add a button component to the textInput component
  - When that button is clicked, it should clear the value of the input (eg. clear the [vue-model](https://vuejs.org/v2/guide/forms.html))
  - The button should only be visible when the input is empty. To do this, I would put a wrapper div around the input & button so they are siblings to each other. Then you can absolutely position the button so it's on top of the input. You can use the `:empty` [pseudo class](https://developer.mozilla.org/en-US/docs/Web/CSS/:empty) on the input to modify the visibility of the button.

Happy to talk through the label + button, and the input clear button next week if you want to pair on it. :) 

Thanks so much for your hard work on this!

jnschrag:
These inputs have been implemented in a different way as a result of work done on the timeline, so I'm going to close this. But thank you so much for your work on this, Jonathan.

PR #78, opened 2020-11-19
Show satellite positions tied to orbits

samadd:
Uses Cesium ComputedPositionProperty to derive an interpolated path for given timeframe and location for any given time point

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,5 +1,6 @@
 {
   "semi": false,
   "arrowParens": "always",
-  "singleQuote": true
+  "singleQuote": true,
+  "trailingComma": false
 }
--- a/.prettierrc
+++ b/.prettierrc
@@ -2,5 +2,5 @@
   "semi": false,
   "arrowParens": "always",
   "singleQuote": true,
-  "trailingComma": false
+  "trailingComma": "none"
 }
--- a/components/visualizer/CesiumViewer.vue
+++ b/components/visualizer/CesiumViewer.vue
@@ -52,7 +52,7 @@ export default {
       alpha: 1,
       brightness: 1,
       contrast: 1,
-      SimInt: 90 * 60,
+      SimInt: 12 * 60 * 60, // 12 hours
       SimStart: null,
       SimStop: null
     }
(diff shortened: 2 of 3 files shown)


.prettierrc
```diff
@@ -1,5 +1,6 @@
 {
   "semi": false,
   "arrowParens": "always",
-  "singleQuote": true
+  "singleQuote": true,
+  "trailingComma": false
```
jnschrag, line 5:
```suggestion
  "trailingComma": none
```

I got a warning that "false" was deprecated, so switching it to "none" instead.

components/visualizer/CesiumViewer.vue
```diff
@@ -103,10 +116,7 @@ export default {
       // For each object, calculate its position & orbit
       // Calculations pulled from: https://github.com/ut-astria/AstriaGraph/blob/master/main.js & https://github.com/ut-astria/AstriaGraph/blob/master/celemech.js
 
-      console.log(this.satellites)
-
-      for (let i = 0; i < this.satellites.length; i++) {
-        const sat = this.satellites[i]
+      this.satellites.forEach((sat, i) => {
```
jnschrag, line 119: I think we probably want to keep this as a regular `for` statement. In all of the bench tests I've seen, `forEach` is a lot slower in comparison. Normally I don't think we have enough data that it really matters, but I think it might be beneficial here.

jnschrag (changes requested):
This is really great, @samadd! I'm so excited that it finally works, and I appreciate you going in and cleaning it up. Glad to see that you were able to reduce a lot of the redundancy I had here. I've asked Thomas to take a look at it from a calculations perspective, but this looks really good to me minus a couple of comments I left.

.prettierrc
samadd: Yep, though "none" needs quotes around it.
```suggestion
  "trailingComma": "none"
```

components/visualizer/CesiumViewer.vue
samadd: Curious - I initially did a quick one-off measure myself using performance.now() -

Time taken with forEach: 1567.744999999879ms
Time taken with for - let: 1562.150000000198ms

... which was quite close so did a few more runs and took an average:

Time taken with forEach: 1578.45199999992
Time taken with for - let:  1622.82399999995

Which is of course the opposite...
That was only 5 runs each, so could be an anomaly, so I'm prepared to be contradicted with more robust metrics on this!

I tend to prefer the functional over imperative style of iteration. While here the difference is fairly trivial, my view is that it is cleaner (no need to assign 'sat' and look up the array index) and in theory allows for the straightforward removal of that inline code to a separate method, which in theory moves toward a more testable code base.

My tests were run in 64-bit Chrome 86 - it's possible that other environments yield different results, but I suspect the overheads that the Array methods may have incurred at one time may no longer be the case in the current version of the V8 engine.

Anyway, there's my defence :)

jnschrag: I'm typically a fan of functional style too, so I'll happily accept that defense. I imagine it matters a whole lot less these days as browsers continue to improve. If we notice any slowness, it's a fairly straight forward thing to change in the future if we need to. :)

jnschrag (approved):

jnschrag:
I'm going to merge this in. I know Thomas was still thinking about the time interval, but I think we can make changes if needed. Thanks so much, Sam!

PR #79, opened 2020-12-01
Show 24 hours, adjust time step for each satellite

samadd:
Adjusts time range to span +- 12hours, adjusts time increment in orbit point calculations based on initial measure of radius and velocity, and adds interpolation setting to sampledPositionProperty


jnschrag (approved):
This looks good to me, Sam. Thank you!

PR #80, opened 2020-12-04
Update satellite.js to consume API using date

samadd:
Satellite store is initialised with a target date set to the appropriate date for the current point in time. The Timeline component will be responsible for updating the targetDate, and triggering a new store dispatch for satellites. Please also confirm that it is safe to remove the commented out early return in getSatellites - unless I'm missing something, that would prevent any API calls being made after the first one?


jnschrag (approved):
This looks great, Sam. Good catch on the commented out line - that was an artifact from copying from the `analysis.js` store where an early return is necessary. We do want to be able to update the store, so good call on removing it.

PR #81, opened 2020-12-10
Feat/timeline component

samadd:
Here is the start of the timeline component. The only thing it really does just now is allow date selection and trigger the entities in the Cesium component to be recalculated. Scale and Speed drop-downs are there but aren't hooked up yet (and so the standard Cesium timeline is still enabled)
I set the max-width on the container css to be auto instead of 1000px - feel free to revert that but since the Figma showed a full-width display I went with that.

### DIFF
--- a/assets/css/abstracts/_mixins.scss
+++ b/assets/css/abstracts/_mixins.scss
@@ -50,6 +50,7 @@
 @mixin vw100() {
   width: 100vw; // Fallback for old browsers
   width: calc(100vw - var(--scrollbar));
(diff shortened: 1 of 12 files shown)


jnschrag (approved):
Thanks so much for this Sam. It looks really good! I added a commit to change the layout of the visualization so it would be full width and not use the page layout used elsewhere on the site.

I'm going to go ahead and merge this in, but next step will be to hook up the timeline to the visualization. I sent an email about the API structure - I'll get that pushed once you've taken a look at the structure.

PR #82, opened 2020-12-11
Feat/visual filters

jnschrag:
@samadd This is the first pass at creating a basic filtering system. Clicking on the "filters" button will take you to a form that lets you filter the satellites. Right now there's only two buttons: the "apply" button filters to only US satellites and the reset button.

While putting this together I realized that we needed to reorganize how the data was being stored in Vuex to make it usable (filtering on all the satellite data was super slow). I've split the data out into two properties in the satellite store: `satellites` and `activeSatellites`. `satellites` is an object that stores all of the information about the satellites, keyed by the `catalog_id`. `activeSatellites` is an array of the `catalog_id`s of all of the satellites that are active according to the filter.

To accommodate this change, I updated the Cesium Viewer to loop over the array of active satellites, and then get the necessary orbital information from there. Per my earlier change to the API so we can return multiple orbits per satellite, I also temporarily set `elems` equal to the first orbit for each satellite.

I wanted to flag this for you in a PR before I merge it in since it's a pretty substantial change. Let me know if you want to talk through anything!

### DIFF
--- a/assets/css/abstracts/_functions.scss
+++ b/assets/css/abstracts/_functions.scss
@@ -1,3 +1,4 @@
+// @use 'variables' as *;
 @use 'sass:math';
 
 /*=================================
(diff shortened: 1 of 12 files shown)


samadd (approved):
All fine with me - good to go.

PR #84, opened 2020-12-15
Select day / week timescale and propagate orbits accordingly

samadd:
This allows you to select a timescale (presently just Day or Week) and calls the API with the startDate and endDate parameters. The orbital path for each satellite is then compiled from each orbital element provided.


jnschrag (approved):
This looks great, Sam. Thank you!

PR #85, opened 2020-12-18
Timeline play pause speed

samadd:
Refrains from trying to calculate orbits for orbitless-satellites, uses more reliable Cesium Timeline insertion method, fixes API call / path rendering bugs and performance issues.


jnschrag (approved):
@samadd I'm gonna merge this in, but could you add some comments to the cesium-service.js file to indicate what is happening there? I'm not entirely sure how services work and would love to learn!

PR #88, opened 2020-12-28
Make datepicker transparent, showing current timeline date

samadd:
The datepicker input is now opacity: 0 and absolutely positioned over a div showing the current date on the Cesium timeline. This is kept in sync using the Cesium clock ontick event. There are some side-effects, in that the date calendar input is offset to the date picker button, but I figured since you'd be doing a lot of css changes anyway, this is something you'd fix / decide on later.


jnschrag (approved):
This is great, thank you Sam! I'll fine tune this as I turn to styling the rest of the timeline. :)

PR #91, opened 2021-01-07
Add 10% buffer to avoid rounding cut-off

samadd:


jnschrag (approved):
This works great! Thanks so much, Sam!

PR #93, opened 2021-01-08
closes #92 role='list' added to Navigation.vue and SocialMedia.vue to correct spacing.

### LINKED ISSUE #92: Add role="list" to list items in footer
In the site footer (#10), add `role="list"` to the navigation & social media share components, which will reset the `padding` to 0. It should also be added to the social share component (#14).

Screen Shot 2021-01-08 at 1 34 48 PM

Screen Shot 2021-01-08 at 1 37 54 PM

Jonathan-J-Choi:


jnschrag (changes requested):
This is great! Can you also add it to the social media component that is at the the top of the page/in the header?

jnschrag (approved):
Thanks so much!

PR #94, opened 2021-01-08
Add track orbit functionality to filter list

jnschrag:
When clicking on the "orbit" button in the filter results list, it should set that satellite as the "tracked" satellite, zooming you into its orbit. It is also set as the selected entity. 

Screen Shot 2021-01-08 at 4 28 14 PM

I _think_ I set up the service correctly, but I did notice that if I clicked around on different satellites, the Cesium viewer started to freak out a bit and I was getting a lot of "requestAnimationFrame handler took Xms" warnings in my console, even if it wasn't playing. I'm not sure if that's related to this at all, but wanted to flag for you in case it was.


samadd (approved):
Changes look good to me - I don't see how anything here would relate to the console error you mentioned.

PR #95, opened 2021-01-11
Bug/earth does not rotate

samadd:
Two issues identified and fixed:
1. The camera was in fact using ECF, and following the Earth. Solution was to continually adjust the camera position as per this SO thread: https://stackoverflow.com/questions/50682925/cesium-earth-show-satellites-in-eci-coordinate-system
2. The satellites themselves were going roughly twice as fast as they should have done. This seems to be fixed by adjusting the reference frame from FIXED to INERTIAL in the SampledPositionProperty.


components/visualizer/CesiumViewer.vue
```diff
@@ -110,6 +110,19 @@ export default {
         }
         entity.path.show = true
       })
+
+      viewer.scene.postUpdate.addEventListener((scene, time) => {
```
jnschrag, line 114: Can you add a brief comment to the code explaining what this does?

components/visualizer/CesiumViewer.vue
```diff
@@ -308,26 +321,45 @@ export default {
...
         interpolationAlgorithm: Cesium.LagrangePolynomialApproximation,
         interpolationDegree: 1
       })
 
-      const duration = this.SimInt * 1.01
-      const orbitalFragmentSize = duration / orbits.length
+      const duration = this.SimInt
+      const numDays = duration / 86400
+
+      const paddedOrbits = []
```
jnschrag, line 335: Can you add a brief comment to the code explaining what this does?

jnschrag (changes requested):
This looks good to me, Sam. I left a couple of comments asking if you could leave a code comment explaining what two sections do. It makes sense to me now, but I imagine in 6 months I'm gonna have no idea why we had to implement those changes. :)

I'm also still getting a really bumpy experience. Even playing just over a single day it's starting/stopping every half a second or so. Do you think this is something that we could fix?

jnschrag (approved):

jnschrag:
This is looking really good, @samadd! I'm noticing a little bit of a hiccup still when you play at the fastest speed with a week or more's worth of data, but it's a lot better than it was. I'll follow up over email, but I'm going to go ahead and merge this in.

PR #101, opened 2021-01-25
Add teardown handlers for Timeline and CesiumViewer

samadd:
The Cesium Viewer.destroy method tries to remove the Timeline DOM element, but as we've moved it, it's not where expected. So We have to be sneaky and wrap the viewer.destroy method and re-insert the timeline DOM element back to where it should be. We also need to deregister the cesium instance, and remove component class scope references to Cesium and the viewer in CesiumViewer when the component instances are destroyed.


jnschrag (approved):
Looks good, thank you!

PR #104, opened 2021-01-27
Reduce path point density to improve playback performance

samadd:
After battling with various approaches to try and show positions and paths that were both non-janky and correct, I found a thread on the Cesium forum that pointed to the path definition as a potential area to look at. The solution to the performance issue wasn't actually explained on the thread, but I thought it worth dialling down the resolution property, and it seems to have made a big improvement to performance as far as I can see.


jnschrag (approved):
This looks great to me. What a bizarre "bug". Thanks for digging into it!

PR #112, opened 2021-02-03
Working on formatting the data for the glossary

jdromero88:


assets/css/pages/post-content.scss
```diff
@@ -8,19 +8,49 @@
     padding: 0 calc(50% - #{$size__content-max-width} / 2);
   }
 
+  div {
+    &:not(:last-child)::after {
+      content: '';
+      display: flex;
+      max-width: $size__line-max-width;
```
jnschrag, line 15: See my other comment about using a border for this, but separately - I'm not sure why `$size__line-max-width` or `$size__alphabet-max-width` needed to be defined. We want our stuff to be fluid & responsive, so if there's a reason for this, let me know so we can discuss.

jnschrag (changes requested):
Overall, really nice job with this! It definitely functionally works, so all of this feedback is just doing the polishing to get it right & optimize it.

## General
- Glossary specific CSS should go in a new file `css/pages/glossary.scss` and then imported in `glossary.vue`. The styles for this page should not override anything that was done for the default typography for the `post-content.scss`.
- I would go through and add classes to the different elements so you can select them specifically. You have to be really careful when selecting general elements like `div` and `section` because they're used in so many places.

## Page Header
- Letters that have terms associated with them in the jump links at the top should have `$color-white-180`. My recommendation is to do something like this:
```
<li v-for="letter in alphabet" :key="letter">
  // if found letters includes letter
  <a :href="`#${letter}`" :aria-label="`Jump to ${letter} words`">{{ letter }}</a> // link
  // else
  {{ letter }} // no link
</li>
```
- Instead of switching between grid & flex, I would recommend using `display: flex;` on the navigation list. This might mean that the # of letters is uneven, but it guarantees that it will wrap no matter the screen size. I'm changing this from what the mockup calls for.

## Definition List
Overall this section looks really good! Just a couple of FYIs:
- The terms should be sorted in alphabetical order within a section. It doesn't look like they have a consistent ordering to them right now.
- I think instead of using `layout-glossary` here, I would use `glossary` as your "block" in the CSS names (ie. `glossary__title`, `glossary__item`, etc.)
- The further reading will actually go inside of the `<dd>` tag, so it'll look something like this:
```
<div v-for="term in termsByLetter(letter)" :key="term.slug">
  <dt class="layout-glossary__title">{{ glossary[term].title }}</dt>
  <dd class="layout-glossary__item">
    {{ glossary[term].short_definition }}
    <div>Further Reading</div>
    Further reading content here.
  </dd>
</div>
```
- Looks like I was too quick to review this morning on our call! The dividing line between terms (eg. between Apogee & Alternate Name) could be done with just a border in CSS. No need for a pseudo element here. It should not be an `<hr>` like I told you earlier.
  - The lines between the different letter sections (eg. between A & C) could be an `<hr>` but let's leave it as is. Apologies for the confusion!
- Double check the font styles on the "back to top" text. The size & the capitalization look off to me. I would give it it's own class so you can target those links specifically.

## JavaScript
Overall I think you did a nice job with the JS. I've got one suggestion that I think you should make, and then a separate one to show you a different approach to solving the overall problem of creating a list of terms grouped by their letter.

For the `foundLetters()` computed property, I think you could use [Sets](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set) to turn this into a clean one-liner.

```
foundLetters() {
  // map & slice to get the first letter of each term. Then transform array into a set & back to a set to remove the duplicates.
  return [...new Set(this.terms.map((term) => term.charAt(0)))].sort()
},
```

If you want to combine the filtering the found letters & grouping them, you could do it like this. We'll be making a computed property called `groupedTerms`. The primary difference between this implementation & yours is that we're only looping over our list of terms once and we're presorting them into their correct places. However, there's no real optimization benefit for one over the other with the scale we're working with

HTML
```
<section v-for="letter in foundLetters" :id="`${letter}`" :key="letter">
  <h2 :id="letter">{{ letter }}</h2>
  <dl>
    <div v-for="term in groupedTerms[letter]" :key="term">
      <dt class="layout-glossary__title">{{ glossary[term].title }}</dt>
      <dd class="layout-glossary__item">
        {{ glossary[term].short_definition }}
        <div>Further Reading</div>
        Further reading content here.
      </dd>
    </div>
  </dl>
  <a href="#top">
    <Icon id="arrow-up" class="icon" name="arrow-up" />
    Back to Top
  </a>
</section>
```


JavaScript
```
computed: {
  foundLetters() {
    return Object.keys(this.groupedTerms).sort() // Objects are not inherently ordered, so we do want to do another sort just to make sure our letters are in the right order.
  },
  groupedTerms() {
    let groupedTerms = {}
    // For each term, get the first letter & capitalize it. If it doesn't already exist in our groupedTerms object, add it. Then add each term as an element to the letter's array so we can loop over it later. Because we pre-sorted the terms, we know they'll be add to the array in alphabetical order.
    this.terms.forEach((term) => {
      let firstLetter = term.charAt(0).toUpperCase()
      if (!groupedTerms[firstLetter]) {
        groupedTerms[firstLetter] = []
      }

      groupedTerms[firstLetter].push(term)
    })

    return groupedTerms
  },
  terms() {
    // Make sure our terms are in alphabetical order from the beginning so we don't have to sort them later.
    return Object.keys(this.glossary).sort()
  },
  ...mapState({
    glossary: (state) => state.glossary.glossary
  })
}
```

assets/css/pages/glossary.scss
```diff
@@ -0,0 +1,120 @@
+a {
+  @extend %font-ui-heading-post-block-title;
```
jnschrag, line 2: I'm not sure why this is being included if you're overwriting the font size & line height down below?

assets/css/pages/glossary.scss
```diff
@@ -0,0 +1,120 @@
...
+  margin-left: rem(12px);
+}
+
+ol {
+  display: flex;
+  flex-direction: row;
+  justify-content: center;
+  margin: 0;
+  padding: rem(13) 0;
+  @extend %font-ui-heading-post-block-title;
```
jnschrag, line 23: Same question here. I think you might want to try `%font-ui-text-base-good` instead.

assets/css/pages/glossary.scss
```diff
@@ -0,0 +1,120 @@
...
+  @extend %font-post-heading-h2;
+  margin-top: rem(54px);
+  margin-bottom: rem(12px);
+
+  color: $color-black-190;
+  line-height: rem(48px);
+  text-transform: uppercase;
+}
+
+div .glossary:not(:last-child) {
```
jnschrag, line 83: You don't need `div` as a part of the selector here. I would also make the class name a little more descriptive, like `glossary-term`, since it's describing a specific element on the page, not the entire page.

pages/glossary.vue
```diff
@@ -1,25 +1,121 @@
...
+      return groupedTerms
+    },
+    terms() {
+      return Object.keys(this.glossary).sort()
+    },
+    ...mapState({
+      glossary: (state) => state.glossary.glossary
+    })
+  },
+  methods: {
```
jnschrag, line 108: You can remove the methods section since you're no longer using any of them.

jnschrag (changes requested):
This is a lot closer! I left a couple of line items to resolve. The other big thing I would suggest is making sure that you're putting your selectors inside of a specific class that is specific to the glossary page & content.

For example, the selectors `li + li` and `a` will apply to **any** list item or link on the page, even if they're in other components. That isn't what we want. Depending on the context, either use a class name to select specific elements, or use nested selectors to get more specific (eg. `.alphabet-nav a` to style the links in the navigation at the top).

assets/css/pages/glossary.scss
```diff
@@ -1,120 +1,117 @@
...
+  li + li {
+    margin-left: rem(12px);
+  }
 
-  &:not(:last-child)::after {
-    content: '';
-    position: absolute;
-    right: 0;
-    left: 0;
+  ol {
```
jdromero88, line 6: The current line height when inspecting on the browser is this:
```line-height: 1.29;```
in the mock up is asking for ```18px``` this is why I was adding the line height like this ```line-height: rem(18px);```
so when is converted we have: ```line-height: 1.125rem```
I just wanted to clarify when is okay to override the line-height?

jnschrag: Good question! In this case, we don't want to override it. Typically when a font size or line height has a discrepancy between what is in the placeholders file & what the Figma mockup calls for, it's better to default to the placeholders file, even if it's slightly different from Figma. In this case, the values are really similar, so it's likely that when the Figma files were being cleaned up, that font style was overlooked. So please don't override the line height in this case. :)

jnschrag (approved):

PR #114, opened 2021-02-04
Zoom to and track selected entity, fix end of path

samadd:
Set tracked entity on viewer.selectedEntityChanged event - this negates the need to set tracked entity in the panel control events, however I've just commented those out for now, in case you want to distinguish between selecting and tracking an entity (in terms of the Cesium Viewer). Selected is when the green tracker is over the entity, tracked is when the camera follows the entity.
Also fixes the wonky straight line on the paths by just adding a full extra cycle on to the orbit calculations.
I've increased the path resolution a bit, to benefit the view at close zoom levels.


jnschrag (approved):
Thanks so much, Sam!

PR #115, opened 2021-02-09
Feat/controls help

### LINKED ISSUE #49: Component: Help Modal
This modal includes information about how to use the visualizer. It makes use of the modal (#48) & tabs (#36) components.

- [ ] Title: `Help`

## About the interface
- [ ] Filtering
- [ ] Pinning Objects

## Changing the View
- [ ] Pan
- [ ] Zoom
- [ ] Rotate
- [ ] Tilt

Jonathan-J-Choi:
closes #49


jnschrag (changes requested):
Overall, I think this is a really solid start. I resolved the merge conflicts, so make sure you pull down the branch. You'll also need to run `npm install` again.

- I'm not sure if it's just me, but the buttons to switch between the two tabs aren't styled. They won't inherit the styles from anywhere else, but you can look at the Details Panel' CSS for reference on how to style them.

Screen Shot 2021-02-09 at 11 36 07 AM

- Remember, `h#` tags have semantic meaning and must occur in a hierarchy. It isn't appropriate to use an `<h4>` tag here because there's no `<h2>` or `<h3>` tag before it.
- My recommendation for the HTML for this is to actually use a [definition list](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dl), where the terms are the headings (eg. Pan, Zoom, etc.) and each term has two definitions, one for desktop & one for mobile. So it would look like this:

```
<dl>
  <div>
      <dt>Pan</dt>
      <dd>Drag one finger</dd>
      <dd>Left click + drag</dd>
  </div>
   <div>
      <dt>Zoom</dt>
      <dd>Drag one finger</dd>
      <dd>Left click + drag</dd>
  </div>
</dl>
```

- You don't need to change this, but this would have been a good opportunity to use nested Sass to reduce some of the repetition, like this:

```
.help {
  &-panel__content {}
  &-title {}
}
```
- There's two CSS selectors for `.help-panel__content`.
- Instead of selecting by ID all of the different icons, you should be able to use a nested selector to get them. Eg. `.help-panel__content .icon` (since no other icons are used in in the content).

assets/css/components/help-panel.scss
```diff
@@ -0,0 +1,91 @@
+.help {
+  &__panel-content {
+    margin-bottom: 44px;
```
jnschrag, line 3: Can you make all of the margin & padding values use `rem` units? We want to make sure it resizes alongside the text.

assets/css/components/help-panel.scss
```diff
@@ -0,0 +1,91 @@
...
+    display: grid;
+    grid-gap: 16px;
+    grid-template-columns: 1fr 1fr;
+    margin-bottom: 32px;
+  }
+
+  &__icon-layout {
+    display: grid;
+    grid-gap: 8px;
+    grid-template-columns: 0.25fr 1fr;
```
jnschrag, line 29: Unfortunately, partial fr units aren't valid. Try `1fr 3fr` instead for the same ratio.

assets/css/components/help-panel.scss
```diff
@@ -0,0 +1,91 @@
...
+    border-radius: 0;
+
+    &:hover {
+      color: $color-white-100;
+    }
+  }
+}
+
+.help-panel__header {
+  max-width: 290px;
```
jnschrag, line 61: Why did you set a max width?

assets/css/components/help-panel.scss
```diff
@@ -0,0 +1,91 @@
...
+}
+
+#view-activator,
+#about-activator {
+  max-width: 138px;
```
jnschrag, line 66: Same question - why did you set a max width?

assets/css/components/help-panel.scss
```diff
@@ -0,0 +1,91 @@
...
+  background: $color-slate-700;
+  border: 0;
+  border-radius: 0;
+
+  &:hover {
+    color: $color-white-100;
+  }
+}
+
+.help__icon-layout .icon {
```
jnschrag, line 79: I would nest this further up where you reference `.help__icon-layout`

jnschrag (changes requested):
Thanks for looking at this! It's a lot closer, but there's still some details that were missed. Take a look at the tab buttons again - they're missing the white underline & the color of the inactive tab isn't correct. See the previous screenshot for reference. It should look & feel exactly like the buttons on the details pane (accessible by clicking on a satellite).

Screen Shot 2021-02-11 at 10 08 53 AM

- The close button should actually close the panel. Here's some [documentation](https://a11y-kit.com/vue-storybook/?path=/story/tabs--outside-control) on how to do that. You can also look in the `FilterTab.vue` component to see how it works on the left side panel.

You don't need to change this, but I wanted to show you how you could simplify the markup a little bit to do the two column grid.

**HTML**
This removes the `help__layout` div.
```
<div class="help__term">
  <dt class="help__title">Pan</dt>
  <dd class="help__icon-layout">
    <Icon id="pan-hand" name="pan-hand" /><span
      class="help__icon-desc"
      >Drag one finger</span
    >
  </dd>
  <dd class="help__icon-layout">
    <Icon id="pan-mouse" name="pan-mouse" /><span
      class="help__icon-desc"
      >Left click + drag</span
    >
  </dd>
</div>
```

**SCSS**
```
.help__term {
  display: grid;
  grid-template-columns: repeat(2, 1fr);
  column-gap: rem(16);
}

.help__title {
  grid-column: 1 / -1; // this will cause the title to span both columns
}
```

jnschrag (approved):
Really nice job, @Jonathan-J-Choi!

I took a look at the vue-tippy repo and found these two issues [one](https://github.com/KABBOUCHI/vue-tippy/issues/78) & [two](https://github.com/KABBOUCHI/vue-tippy/issues/38) which helped me figure out how to add the close functionality to the panel. :)

PR #119, opened 2021-02-16
Feat/further reading

jdromero88:


jnschrag (changes requested):
This is looking good, @jdromero88! I made an update to the packages file so it pulled in the most recent packages for the site. Make sure you pull the branch back down & run `npm install` before you work on it again.

In addition to the one comment I left, please take a look at the typography of the further reading section again. I believe it uses a smaller font than the definition does. Can we also increase the space between the further reading section & the definition to `24px`? Thank you!

jnschrag (approved):

PR #120, opened 2021-02-17
Bug/earth time

jnschrag:
A WIP attempt at fixing the satellite's locations, which seem to be off from the earth's location by about 9-10 hours.

## Issue
The satellite's locations relative to the rotation of the earth is incorrect. There is a gap in the satellites around the equator that should be located over the Pacific Ocean. See Celestrak & Astriagraph as an example:

Screen Shot 2021-02-17 at 2 47 04 PM
Screen Shot 2021-02-17 at 2 46 52 PM

In our implementation, the gap appears over the Indian Ocean.
Screen Shot 2021-02-17 at 2 49 03 PM

## Possible Solutions
This fix is incomplete, but it does a couple of things:
- On page load, get the current date **and** time with `new Date()`. It no longer resets the hours to 0.
- On the Cesium viewer, do not reset the hours to 0 when working with the epoch.

This results in the gap shifting so it is over the Pacific, but it is still in the wrong location (it's located over Asia as opposed to the ocean).
Screen Shot 2021-02-17 at 2 52 26 PM

The gap stays in the same place when using the next/previous buttons or scrubbing the timeline. However, when selecting a specific date from the date picker, the earth rotates so the camera is focused on the Atlantic/Africa/Europe, "moving" the gap in the satellites.
Screen Shot 2021-02-17 at 2 54 57 PM

## Takeaways
There seem to be a couple of things at odds here:
- The location of the satellites
- The rotation of the earth at a particular date/time
- Where the camera focuses on when you select a new target date

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -105,3 +105,5 @@ amplifyconfiguration.json
 amplify-build-config.json
 amplify-gradle-config.json
 amplifyxc.config
+# Local Netlify folder
+.netlify
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -4,5 +4,7 @@
   "vetur.format.defaultFormatter.html": "prettier",
   "editor.codeActionsOnSave": {
     "source.fixAll.eslint": true
-  }
+  },
+  "liveServer.settings.multiRootWorkspaceName": "satellite-dashboard",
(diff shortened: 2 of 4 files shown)


jnschrag:
Closing this as it's been resolved by #130.

PR #130, opened 2021-02-25
Remove CRFtoTRF transformation

samadd:


samadd:
The CRFtoTRF transformation seems to have been the source of the erroneous offset on the satellites. Presumably the use of Cesium.ReferenceFrame.INERTIAL renders it unnecessary.

jnschrag (approved):
This looks great to me, thanks so much!

PR #132, opened 2021-02-28
Added sort function

jdromero88:


components/visualizer/details-panel/Events.vue
```diff
@@ -162,6 +174,17 @@ export default {
   methods: {
     updateMaxDistance() {
       this.appliedMaxDistance = this.maxDistance
+    },
+
+    updateCurrentSort(value) {
```
jnschrag, line 179: Can you explain why this is a method? If we're using `v-model` on the dropdown, it will update `currentSort` automatically, and the `filteredEvents` computed property will update automatically as well.

components/visualizer/details-panel/Events.vue
```diff
@@ -151,9 +153,19 @@ export default {
...
+      let selected = this.currentSort
+      let events = this.formattedEvents.filter(
         (d) => d.min_distance_km <= this.appliedMaxDistance
       )
+
+      if (selected === 'time_of_close_approach') {
+        events.sort((a, b) => new Date(a[selected]) - new Date(b[selected]))
+      }
+
+      if (selected === 'min_distance') {
```
jnschrag, line 165: This should be an `elseif`. I would also recommend setting the sort value to `min_distance_km`, and then updating this so it's `events.sort((a, b) => a[this.currentSort] - b[this.currentSort])`. It's more reusable that way.

components/visualizer/details-panel/Events.vue
```diff
@@ -151,9 +153,19 @@ export default {
       return this.filteredEvents.length
     },
     filteredEvents() {
-      return this.formattedEvents.filter(
+      let selected = this.currentSort
```
jnschrag, line 156: I would recommend against creating another variable and using `this.currentSort` directly.

jnschrag (changes requested):
Really  nice job on this, José! The functionality works great. There's a couple of style things we need to address and then I have some suggestions for refactoring to make this a little bit cleaner:

- The options listed in the sort dropdown are correct, but when you select them, it shows the selected value as opposed to the label. We want the label to be what is selected.

Screen Shot 2021-03-01 at 1 19 00 PM


- Double check the styles in Figma. In this instance, we want:
  - The Total # of results should be on the same line as the select box (I would use `display: flex` on the `.details-events__totals` div)
  - We can remove the arrow from the dropdown

components/visualizer/details-panel/Events.vue
jdromero88: So the reason I did not use the ```else if (condition) { }``` was just because I'm trying to use the coding style of the [line of sight](https://medium.com/@matryer/line-of-sight-in-code-186dd7cdea88).

components/visualizer/details-panel/Events.vue
jdromero88: I used a method because it wasn't updating the ```CloseApproachesList``` component. So my solution was to use a method to update the ```currentSort``` so this way I was like "forcing" the component to update.
After you review I went back to the docs & I found that I could use the ```reduce``` property to achieve the same thing without using any methods. 💯 👍🏽

jnschrag (approved):
Really nice job!

PR #134, opened 2021-03-03
Updated references to get the info

jdromero88:


jnschrag (changes requested):
This looks great! Now that I'm seeing this in context, there's a couple of CSS changes I think we should make:

- On `.post__further-article`, set `align-items: flex-start` and `gap: rem(24)`
- On `.post__further-circle`, set `flex: 0 0 auto`

This will help with the spacing and sizing of the link icon. :)

jdromero88:
I removed all the ```px``` from the ```rem()``` function

jnschrag (approved):
Looks great, thank you!

PR #138, opened 2021-03-08
Gave format to footnote

jdromero88:


pages/analysis/_slug.vue
```diff
@@ -46,24 +46,34 @@
...
-              {{ reading.publication_name }}
-              <span class="post__further-source">{{
-                reading.publication_organization
-              }}</span>
-            </a>
-            <a :href="reading.url" class="post__further-circle">
-              <Icon class="post__further-icon" name="external-link" />
-            </a>
-          </div>
+        <template v-if="post.footnote">
```
jnschrag, line 49: Instead of using `<template>` here, you can apply the `v-if` to the `<ol>` tag directly.

pages/analysis/_slug.vue
```diff
@@ -46,24 +46,34 @@
...
+          <ol class="analysis__notes">
+            <template v-for="(note, index) in footNote">
```
jnschrag, line 51: Same here, you should be able to apply `v-for` directly on the `<li>` tag.

pages/analysis/_slug.vue
```diff
@@ -46,24 +46,34 @@
...
+                v-html="note"
+              ></li>
+            </template>
+          </ol>
+        </template>
+        <template v-if="post.acf.further_reading">
+          <h2 class="post__further-header">
+            Further Reading
+          </h2>
+          <template v-for="reading in post.acf.further_reading">
```
jnschrag, line 64: Same here. The `v-for` can be applied to the div (but this should also be a list too `<ul>` with each entry a `<li>`).

store/analysis.js
```diff
@@ -73,7 +74,8 @@ export const actions = {
             acf,
             categories,
             country,
-            user
+            user,
+            footnote: content.rendered.match(regex)
```
jnschrag, line 78:
```suggestion
            footnotes: content.rendered.match(regex)
```

I'd make this plural because you're expecting to get multiple matches.

jnschrag (changes requested):
Thanks for digging into this! I know that Regex is tough, so I appreciate it! I left a few notes inline and then have these to flag:

- The footnote styles should be included in `post.scss`, not `analysis.scss` (`analysis.scss` is for `/analysis` the archive page)
- Naming things isn't a perfect science & it can be really tricky, but in general, it's advised to use the singular form if you're referring to a single item and the plural form if you're referring to multiples. I would rename `footNote` to `footnotes` accordingly. The camelCase doesn't really make sense here since "footnote" is one word.

jnschrag (approved):

jnschrag (approved):

store/analysis.js
```diff
@@ -75,6 +76,7 @@ export const actions = {
               categories,
               country,
               user,
+              footnotes: formatFootnotes(content.rendered)
```
jnschrag, line 79:
```suggestion
              footnotes: formatFootnotes(content.rendered),
```

PR #142, opened 2021-03-12
Feat/specify satellites time query string

samadd:
Accepts values for satids (comma-separated id list), date (YYYY-mm-dd), time (seconds after midnight on the selected date) and timescale ('day', 'week', 'month')
Also in Cesium Viewer, re-create entities when "de-focusing", otherwise they don't exist and can't be re-shown.

### DIFF
--- a/components/visualizer/CesiumViewer.vue
+++ b/components/visualizer/CesiumViewer.vue
@@ -146,9 +146,9 @@ export default {
         return
       }
       if (this.$route.query.time) {
-        const seconds = parseInt(this.$route.query.time)
+        let seconds = parseInt(this.$route.query.time)
         if (isNaN(seconds)) {
-          viewer.clock.currentTime = this.SimStart
+          seconds = 0
         }
         viewer.clock.currentTime = Cesium.JulianDate.addSeconds(
           this.SimStart,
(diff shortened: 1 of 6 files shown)


components/visualizer/CesiumViewer.vue
```diff
@@ -140,6 +141,24 @@ export default {
...
+    handleTimeQuery() {
+      if (!viewer || !Cesium) {
+        return
+      }
+      if (this.$route.query.time) {
+        const seconds = parseInt(this.$route.query.time)
+        if (isNaN(seconds)) {
+          viewer.clock.currentTime = this.SimStart
+        }
+        viewer.clock.currentTime = Cesium.JulianDate.addSeconds(
```
jnschrag, line 153: Do you need an `else` block here? Otherwise doesn't this line negate anything that happens on line 151?

jnschrag (approved):
This looks great to me, @samadd! I had one question about the logic in the Cesium Viewer, but otherwise I can merge this in.

Saving these strings here for posterity: `?satids=11570,21052,16667&date=2021-02-28&time=86400&timescale=week`

jnschrag (approved):

PR #143, opened 2021-03-12
Add tooltips to Entities

jnschrag:
**Still a WIP**

The default "label" for Cesium entities do not render well - they're very blurry. I had [originally updated](https://community.cesium.com/t/clean-text-labels-with-cesium/12617) the selectionIndicator HTML to show the name of the entity, but since we want to be able to show more than one path at a time, I opted for a solution that would add an HTML element to hold each label.

Sam, I was hoping you could take a look at the code in CesiumViewer.vue and let me know what you think? Everything functions the way that it should, but I'm a bit concerned that it's not particularly performant, so I'm interested in your thoughts. I wasn't sure if the position calculations should happen in the `viwer.scene.postUpdate` section as opposed to where I've put them.

### DIFF
--- a/components/visualizer/CesiumViewer.vue
+++ b/components/visualizer/CesiumViewer.vue
@@ -101,7 +101,7 @@ export default {
       SimStop: null,
       showSunlight: true,
       satellitesHaveLoaded: false,
-      selectionIndicator: false,
+      selectionIndicator: true,
       defaultZoomAmount: 15000000,
       defaultPosition: {
         lng: -80,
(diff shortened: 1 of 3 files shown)


components/visualizer/CesiumViewer.vue
```diff
@@ -157,26 +159,106 @@ export default {
...
+      viewer.container.appendChild(labelContainer)
+
+      let entityLabel = document.createElement('div')
+      entityLabel.classList.add('cesium__entity-label')
+
+      let highlightedEntities = new Set()
+
+      // Every animation frame, update the HTML element position from the entity.
+      viewer.clock.onTick.addEventListener(function(clock) {
+        var position3d
```
samadd, line 182: Minor - ideally stick to either const or let and avoid var for declaration.

samadd (approved):
Looks fine - although there's a temptation to go for a declarative approach in rendering html labels (since it seems wrong to be manipulating the DOM directly in a Vue app), I think that that would incur an overhead, so this imperative, direct approach is the best way to go with this. I haven't noticed any performance problems when trying it out, either.

jnschrag:
Thanks for reviewing, Sam! Much appreciated. :)

PR #144, opened 2021-03-23
Bug/camera jump on selection change

samadd:
This branch fixes the strange path distortion by changing viewer.selectedEntity on target press in order to trigger the camera positioning logic in the selectedEntityChanged listener.
I've made some additional changes to stop the camera flying away from the scenes of interest at the first opportunity. There are still issues (for example the flash of position change before the camera starts moving) but I think it's going in the right direction.

### DIFF
--- a/components/visualizer/CesiumViewer.vue
+++ b/components/visualizer/CesiumViewer.vue
@@ -212,6 +212,10 @@ export default {
             label = document.getElementById(`entity-${entity.id}`)
           }
 
+          if (!label) {
+            return
+          }
+
           // Moving entities don't have a position for every possible time, need to check.
           if (position3d) {
             position2d = Cesium.SceneTransforms.wgs84ToWindowCoordinates(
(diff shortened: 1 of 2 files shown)


components/visualizer/CesiumViewer.vue
```diff
@@ -240,23 +244,24 @@ export default {
...
-              viewer.selectedEntity = false
-              viewer.trackedEntity = false
-
               highlightedEntities.delete(entity)
               entity.path.show = false
               document.getElementById(`entity-${entity.id}`).remove()
+              viewer.selectedEntity = null
             } else {
               // Update highlightedEntries & add label when the selectedEntity changes to make behavior consistent across viewer & buttons on panels.
+              highlightedEntities.add(entity)
```
jnschrag, line 259: Doesn't this already happen on line 282 inside the `selectedEntityChanged` function? It's a set, so it doesn't really matter, but curious as to why this was moved here.

samadd: Yeah, the reason was that the check in selectedEntityChanged for the size of highlightedEntities needed to take place under certain conditions, so I think I needed to add it there for certain scenarios. My hope, however, is that when looking at the tracking bug you mentioned in your email, I'll be able to tidy up the admittedly confusing logical knots that have emerged here!

jnschrag (approved):
This is fantastic, Sam. It works great and seems to run pretty smoothly. I appreciate you jumping in to solve this logic puzzle!

PR #146, opened 2021-05-04
Closes #145 | Improving Footnotes Functionality

### LINKED ISSUE #145: Improving Footnotes Functionality
## Problem
While we have a basic footnotes functionality implemented ([see here for example](https://satellitedashboard.org/analysis/mev-1-intelsat-901), there's one big change we need to make before we can call this feature "done". In the body text, a superscript number denotes the location of a footnote, which is great. But that note still links to our WPE site, and not to the real site.

## Proposed Solution
We need to modify the footnote link so it directs to the footnotes at the bottom of the post. For example, if this is the default HTML for the footnotes:

```html
<span id="easy-footnote-3-6834" class="easy-footnote-margin-adjust"></span>
<span class="easy-footnote"><a href="https://satdash.wpengine.com/mev-1-intelsat-901/#easy-footnote-bottom-3-6834" title="Ibid."><sup>3</sup></a></span>
```

We would want to change it to this:
```html
<span id="easy-footnote-3-6834" class="easy-footnote-margin-adjust"></span>
<span class="easy-footnote"><a href="#easy-footnote-bottom-3-6834" title="Ibid."><sup>3</sup></a></span>
```

The second part of this is updating the footnote list at the bottom to have the correct ID assigned to it (`easy-footnote-bottom-3-6834` in this example), so that clicking on the link in the body text scrolls the user down the page to the correct footnote.

jdromero88:
Closes #145 Update/footnotes link


jnschrag (approved):
Thanks so much, @jdromero88. This technically works great. I do have one concern, however. And it's that these changes take place on the client. That in and of itself isn't a huge deal, but I'd be curious if we would be able to make these changes during the build process.

When you first set up [the footnotes](https://github.com/CSIS-iLab/satellite-dashboard/blob/main/store/analysis.js#L185), you and @smaraghi had discussed using an HTML parser instead of regex. I would be interested to know if using something like [node-html-parser](https://www.npmjs.com/package/node-html-parser) would work to accomplish both tasks:

1. Pull out the footnotes content from the post content
2. Update the post content footnote links to direct to the footnote anchor link (what you've done in this PR so far).

Looking at the documentation of that HTML parser, I wonder if we could accomplish task 2 like this:

```js
import { parse } from 'node-html-parser';

const content = parse(content.rendered);
const footnotes = content.querySelectorAll('.easy-footnote a'); // should return all footnote links
footnotes.forEach(footnote => footnote.setAttribute('href', ID HERE))
```

This is incomplete code, and would need to be tested. But this might get you started. Can you spend some time this week playing with it? I looked at this [test](https://github.com/taoqf/node-html-parser/blob/master/test/html.js#L303) for reference on updating the attribute.

You can also look at [htmlparser2](https://www.npmjs.com/package/htmlparser2) to see if that's easier to use.

For awareness, I'm pushing to update this to use an HTML parser because it's less fragile than regex, and we are eventually going to need to do something similar to show glossary terms. If we can figure it out for footnotes, we'll be able to use a similar pattern for the glossary.

Please time box yourself on this: don't spend more than 2-3 hours trying to work with the HTML parser (if you've finished your other tasks & you have more time to work on this, you can). Can you reply back to this PR with what you've discovered? Please make sure you comment out the working solution you've submitted now, just so we don't lose it if we decide the HTML parser method doesn't work for this use case.

package.json
```diff
@@ -60,6 +62,7 @@
     "stylelint": "^13.1.0",
     "stylelint-config-sass-guidelines": "^7.0.0",
     "stylelint-order": "^4.0.0",
-    "stylelint-scss": "^3.14.2"
+    "stylelint-scss": "^3.14.2",
+    "typescript": "^4.2.4"
```
jnschrag, line 66: Why was typescript added to this project? That's a really big dependency to add.

package.json
```diff
@@ -52,6 +52,8 @@
     "html-entities": "^2.1.0",
     "husky": "^4.0.0",
     "lint-staged": "^10.0.0",
+    "merge-descriptors": "^1.0.1",
```
jnschrag, line 55: What is this package for? I don't see it used anywhere.

store/analysis.js
```diff
@@ -183,6 +185,31 @@ export const actions = {
...
+  let newContent = parse(content)
+  // Grabs all the spans that have the id that we need
+  let spansID = newContent.querySelectorAll('.easy-footnote-margin-adjust')
+  // Grabs all the spans that contains the a tags
+  let spansAEl = newContent.querySelectorAll('.easy-footnote a')
+  if (spansAEl && spansAEl.length) {
+    spansAEl.forEach((a, i) => {
+      a.removeChild(a.firstChild)
+      a.setAttribute('id', spansID[i].id)
+      spansAEl.forEach((a) => {
```
jnschrag, line 198: Can you explain what is happening on lines 198-200? I don't understand why we're doing another loop over `spansAEL` and not calling `a.set_content(a.__attrs.title)` directly.

store/analysis.js
```diff
@@ -183,6 +185,31 @@ export const actions = {
 }
 
 function formatFootnotes(content) {
+  let formattedFootnotes = null
+  let newContent = parse(content)
+  // Grabs all the spans that have the id that we need
+  let spansID = newContent.querySelectorAll('.easy-footnote-margin-adjust')
+  // Grabs all the spans that contains the a tags
+  let spansAEl = newContent.querySelectorAll('.easy-footnote a')
+  if (spansAEl && spansAEl.length) {
```
jnschrag, line 194: It would be better to do an early return here if these conditions aren't true, rather than putting the bulk of your code inside of an `if` statement.

store/analysis.js
```diff
@@ -183,6 +185,31 @@ export const actions = {
...
+    spansAEl.forEach((a, i) => {
+      a.removeChild(a.firstChild)
+      a.setAttribute('id', spansID[i].id)
+      spansAEl.forEach((a) => {
+        a.set_content(a._attrs.title)
+      })
+    })
```
jnschrag, line 201: You could also populate the contents of `formattedFootnotes` in this loop as well, instead of doing that separately below by doing something like:
```
formattedFootnotes.push({
  id: `${a._attrs.id}-bottom`,
  title: a._attrs.title
})
```

It's a micro-optimization, but it saves you from looping over `spansAEl` again (which is what `.map()` does).

jnschrag (changes requested):
Thanks for looking into this, José! I think we're making some positive steps in the right direction, and I like the HTML parser technique more than the regex. I think it's less brittle overall, and if Easy Footnotes ever updates it's HTML structure, it'll be a lot simpler for us to update too.

I left some inline comments on making specific improvements, but I think the next thing to try to do is update the `content.rendered` with the changes you're making in the `formatFootnotes`. Those changes should update the content so the HTML links to the footnotes is correct, rather than doing it in the `mounted()` hook.

In practice, I think this means that this formatting function will need to return 2 things: the content & the formatted footnotes. So instead of calling it on line 80 like you are now, I would probably do something like, `const formattedContent = formattedContent(content.rendered)` on line 64, before the `return` statement. Then Your `formattedContent` function should return an object that contains the content & footnotes, so you can access them like this: `formattedContent.content` and `formattedContent.footnotes`.

This pattern will set us up well for the other manipulations of the content we're going to need to do for the glossary.

I think we're really close, there's just a bit more to do! You did a great job tackling this task, and I appreciate the work you put into it.

store/analysis.js
jdromero88: That's a great question: so when I was using directly ```a._attrs.title``` I was returning me undefined (no idea why, most likely my PC was just being difficult, or I made a mistake with accessing the key) because now I tried & is just working fine.

Anyways my workaround was by looping in again and there I got access to the ```_attrs``` which now that you mention it makes no sense looping 2 times.

package.json
jdromero88: to be honest I have no idea. I added that because whats throwing me this error:
Screenshot at 2021-05-18 09-49-06

so I just added to the devDependecies to be able to run the project but I forgot to exclude it the package.json from the commit.

Sorry.

package.json
jdromero88: I added typescript because the warnings
Screenshot at 2021-05-18 09-56-04
but I forgot that I didn't supposed to add the package.lock to my commit. Sorry @jnschrag

jnschrag (changes requested):
This is fantastic, José! Really, really well done. I've got two nitpicky, code-style things but otherwise, I think this PR is good to go. Can you make these adjustments and then change this from draft to published? I'll review it one last time and then merge it in. :)

1. Instead of adding `formattedContent` to your post object, I would do something like this instead that keeps the content & footnotes separate.

```
const formattedContent = contentFormatter(content.rendered)
return {
  id,
  slug,
  title,
  excerpt,
  date,
  modified,
  tags,
  content: formattedContent.content,
  authors: coauthors_full,
  meta,
  acf,
  categories,
  country,
  user,
  footnotes: formattedContent.footnotes,
  satellites: getSatellites(acf),
  searchable: searchableContent(title, content)
}
```

You can then reference them as `content` and `footnotes` where you need to.

As a follow up to this, can you make sure that all references to `content.rendered` are updated accordingly to just `content`?

2. While in some code bases, it's common to put constants in all caps, that isn't the style for this project. Can you update `const CONTENT_PARSED` to be `const content_parsed`, and update all references to that variable accordingly?

Thanks so much! This really is great. One of the next things I'm going to ask you to take a look at is the glossary terms, and this implementation is going to make a huge difference in our ability to do that, so nice work!

jnschrag (approved):
Really nice work. :) I added in that merge dependency (I didn't realize it was required for `vue-cesium`). But this is fantastic. Thank you so much!

PR #147, opened 2021-05-04
Added related post

jdromero88:


pages/analysis/_slug.vue
```diff
@@ -104,7 +104,26 @@
         </div>
       </footer>
     </div>
-    <section class="post__related-wrapper"></section>
+    <section class="post__related-wrapper">
```
jnschrag, line 107: We should add a `v-if` statement to this, so this whole block of code is only rendered if we have `relatedPosts`.

assets/css/pages/post.scss
```diff
@@ -351,15 +351,21 @@
...
+      align-items: center;
       width: 100vw;
       height: rem(240);
       overflow: hidden;
       text-align: center;
-      background-color: $color-slate-200;
+      background-color: $color-slate-600;
 
       @include breakpoint('medium') {
+        display: flex;
```
jnschrag, line 365: You don't need to redeclare this, since you've already defined the display as `flex` above.

assets/css/pages/post.scss
```diff
@@ -351,15 +351,21 @@
     }
 
     &-block {
-      display: table;
+      // display: table;
```
jnschrag, line 354: Remove this if you're not using it.

assets/css/pages/post.scss
```diff
@@ -351,15 +351,21 @@
...
+      display: flex;
+      justify-content: center;
+      align-items: center;
       width: 100vw;
```
jnschrag, line 358: You shouldn't need to include a width property here. It'll default to 100% of it's parent container by default (which is 100vw in this case).

assets/css/pages/post.scss
```diff
@@ -351,15 +351,21 @@
...
       height: rem(240);
```
jnschrag, line 359: We don't want to set an explicit height here because we need to account for varying lengths of titles. Instead, you should adding padding to the related post block so it can responsively size based on the content and still have space around the text content. We can also get rid of the `overflow: hidden` declaration.

assets/css/pages/post.scss
```diff
@@ -351,15 +351,21 @@
...
       overflow: hidden;
       text-align: center;
-      background-color: $color-slate-200;
+      background-color: $color-slate-600;
 
       @include breakpoint('medium') {
+        display: flex;
         flex: 1 1 50%;
+        justify-content: center;
```
jnschrag, line 367: You don't need to redeclare these since they're the same as the base styles.

jnschrag (changes requested):
Thanks for taking a first pass at this, José! Overall this is really well done. I have a couple of suggestions for how we can improve this so it's more semantic & consistent with other implementations on the site.

- There are differing opinions on this, but I think it would make sense to use an `<aside>` element instead of a `<section>` element for the related posts wrapper element. Since it's content that is related to the main content of the page, but only indirectly. See [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/aside) for an explanation.
- The related articles need to link to their respective posts. Right now that link is missing. My recommendation would be to wrap the title in a link, and then you can use CSS to extend that link so it covers the whole area of the card. [Here](https://codepo8.github.io/full-card-click/2-clickable-card.html) is an example of that (and [here](https://christianheilmann.com/2020/11/26/back-to-basics-creating-a-clickable-card-interface-in-plain-html-css-and-javascript/) is a link to the full tutorial - ignore the options that go into buttons and other links, our use case is much simpler than that). In our case, we want the link to be around the post title.
- I see that you commented out a line that uses the [PostMeta component](https://github.com/CSIS-iLab/satellite-dashboard/blob/main/components/global/PostMeta.vue) - why didn't you use this? I would recommend using that so the HTML & CSS classes are the same as the other implementations used on the site. It's okay that the date will come before the author information.
- We can improve the flexbox implementation a little bit. In this case, I would actually recommend using 
CSS grid instead of flexbox to lay out the posts next to each other. You can do that with the below code, which says that our columns need to be a minimum of 300px wide. If we don't have enough space for there to be multiple columns at 300px wide, then they should wrap & become rows instead.
```
  &__related {
    &-wrapper {
      display: grid;
      grid-template-columns: repeat(auto-fit, minmax(300px, 1fr));
    }
  }
```

jdromero88:
Hola @jnschrag thanks for the great feedback!

Regarding this:
- "I see that you commented out a line that uses the PostMeta component - why didn't you use this? I would recommend using that so the HTML & CSS classes are the same as the other implementations used on the site. It's okay that the date will come before the author information."

I used the PostMeta component
Screenshot at 2021-05-13 10-30-09

and then I commented it out to follow the figma mockup where the date & author are on a single line.
Screenshot at 2021-05-13 10-31-10

pages/analysis/_slug.vue
```diff
@@ -104,7 +104,24 @@
         </div>
       </footer>
     </div>
-    <section class="post__related-wrapper"></section>
+    <aside v-if="relatedPosts" class="post__related-wrapper">
+      <!-- Here goes the related post -->
+      <template v-for="relatedPost in relatedPosts">
+        <article :key="relatedPost.ID" class="post__related-block">
+          <div>
```
jnschrag, line 111: I think you can remove this div, since it doesn't seem to serve a purpose other than being a container for the post block content, and I think the `<article>` tag does that. The CSS related to `flex` on the `<article>` will need to be modified slightly, but that shouldn't be too much of an issue.

assets/css/pages/post.scss
```diff
@@ -340,26 +340,40 @@
...
+      align-items: center;
+      padding-top: rem(58);
+      padding-bottom: rem(57);
       text-align: center;
-      background-color: $color-slate-200;
+      background-color: $color-slate-600;
 
       @include breakpoint('medium') {
-        flex: 1 1 50%;
         width: 50vw;
```
jnschrag, line 377: You actually don't need to set `width: 50vw` here either. The width of the block is taken care of by the grid styles applied on the `.post__related-wrapper`.

assets/css/pages/post.scss
```diff
@@ -340,26 +340,40 @@
...
 
     &-block {
-      display: table;
-      width: 100vw;
-      height: rem(240);
-      overflow: hidden;
+      display: flex;
+      justify-content: center;
+      align-items: center;
+      padding-top: rem(58);
```
jnschrag, line 371: I would add 16px of padding (converted to rems) on the right & left sides too, just to make sure the text never runs up against the sides.

jnschrag (changes requested):
Thanks for making these updates. And apologies for not being clear: the publish date & author information typography should look the same as what they are in the mockups, they just don't need to be in the same order (it's fine that the date comes first). You can override the styles by using the selector `.post__related-block .post-meta__label` and `.post__related-block .post-meta dd`, which will target the post meta only in the related post block, nowhere else on the page.

jdromero88:
Hi @jnschrag now the only difference are the ```by``` is in lowercase and the month is the short format ```Aug```
Screenshot at 2021-05-18 13-02-36
not sure if this is a issue but definitely is bugging me. jeje

jnschrag (approved):
Nice work, José! That format matches the other places that component is used on the site, so it actually is better for consistency than what is in the mockup. :) Appreciate you working on this!

PR #151, opened 2021-05-21
Buttons with icons more accessible 

### LINKED ISSUE #149: Giving all Icon Buttons an `aria-label` to improve accessibility
While making the site, I did a poor job at making sure that all of the buttons were accessible. Many of them follow this pattern:

```
<Button
   type="button"
   class="modal__btn-close-icon"
   :on-click="close"
>
    <Icon id="close-small" name="close-small" />
</Button>
```

There are 2 main issues with this:
1. There is not descriptive text associated with the button, so a user using a screenreader won't know what this button does.
2. The [`<Icon />` Component](https://github.com/CSIS-iLab/satellite-dashboard/blob/5ae69a67b624c62512b414c7ea4373b16a211be1/components/global/Icon.vue) is focusable, which means the `<svg>` could trap the keyboard focus, and the `title` attribute on the `<Icon />` would be read instead.

## The Fix
All buttons on the site using this pattern need to be updated. There are two main ways to fix these buttons, depending on the use case.

### Icon only button
The above could be changed to this, which adds `aria-label` to the `<Button />` and `focusable="false"` to the `<Icon />`.

```
<Button
   type="button"
  aria-label="Close Modal"
   class="modal__btn-close-icon"
   :on-click="close"
>
    <Icon id="close-small" name="close-small" focusable="false" />
</Button>
```

### Icon + Text Button
For buttons that contain both text & an icon, no `aria-label` is necessary, but the Icon should be hidden by adding `aria-hidden="true"` to `<Icon />`. This tells a screenreader or other assistive technology to skip over it, and is ideal for circumstances like this when the icon is only for visual purposes.

```
<Button
   type="button"
   class="modal__btn-close-icon"
   :on-click="close"
>
    <Icon id="close-small" name="close-small" focusable="false" aria-hidden="true" />
    Close Menu
</Button>
```

More information can be found in [this article](https://www.sarasoueidan.com/blog/accessible-icon-buttons/#technique-3-using-aria-label) by Sara Soueidan.

Jonathan-J-Choi:
Buttons with icons given aria-labels and icon is no longer focusable.  Buttons with icons and text given aria-hidden and focusable false on icons.  Closes #149

### DIFF
--- a/components/global/Header.vue
+++ b/components/global/Header.vue
@@ -1,12 +1,25 @@
 <template>
   <header class="site-header">
     <div class="site-header__masthead">
-      <Button class="btn--menu" :on-click="openNav">
-        <Icon id="menu" class="icon header__menu" name="menu" />
+      <Button
+        aria-label="Open Navigation Menu"
+        class="btn--menu"
+        :on-click="openNav"
+      >
+        <Icon
+          id="menu"
+          class="icon header__menu"
+          name="menu"
+          focusable="false"
+        />
       </Button>
       <div class="site-nav" :class="navIsOpen ? 'site-nav--is-visible' : ''">
-        <Button class="btn--icon btn--close" :on-click="closeNav">
-          <Icon id="close-large" name="close-large" />
+        <Button
+          aria-label="Close Navigation Menu"
+          class="btn--icon btn--close"
+          :on-click="closeNav"
+        >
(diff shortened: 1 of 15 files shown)


jnschrag (approved):
Really nice work on this, Jonathan! Appreciate you going through this for me. :)

PR #152, opened 2021-05-24
Closes #148 | Post: Updated Section

### LINKED ISSUE #148: Post: Updated Section
Analysis posts need to accommodate the ability to include an "Updated" section at the bottom. This is so the research team can add updates to posts about a particular event or satellite, and can clearly indicate that it's been updated.

Screen Shot 2021-05-18 at 10 24 35 AM

[View](https://www.figma.com/file/XzGzlXpNmeB3PzTNfzk4ro/Satellite-Dashboard%2C-handoff?node-id=113%3A1401) the above in Figma.

To create this section, we'll need to create a Gutenberg block that the program can add to the end of their posts. The block should be called "Last Updated Section", and should have a date field. The program will then implement their content like normal using regular blocks.

The block styles should include the styling of the date field (typography) and an `<hr />` tag to semantically separate this part of the post from the other content.

## To Complete:
- [ ] Create Gutenberg Block
- [ ] Style Block (can be added to one of the test posts)

jdromero88:
Closes #148 
Hi @jnschrag 
You can test the new feature in this link[http://localhost:3000/analysis/mev-1-intelsat-901](http://localhost:3000/analysis/mev-1-intelsat-901)

Also looks like the class ```.entry-content``` should have a ```background-color: #F2F2F2;``` but please correct me if I am wrong?
P.S. Because the background is white I left the Updated Section bg color also white.

### DIFF
--- a/assets/css/pages/post.scss
+++ b/assets/css/pages/post.scss
@@ -172,30 +172,6 @@
     }
   }
 
-  // Last updated
-  &__last-updated {
-    padding: 0 calc(50% - #{$size__content-max-width} / 2);
-    background-color: $color-white-100;
-    @include full-width-background($color-white-100, true);
-
-    @include breakpoint('medium') {
-      &::before,
-      &::after {
-        display: none;
-      }
-    }
-
-    &-date {
-      @extend %font-ui-text-sm-oxygen;
-      color: $color-black-160;
-    }
-
-    &-separator {
-      margin-bottom: rem(48);
-      color: rgba(0, 0, 0, 0.2);
-    }
-  }
-
   // Further reading section
 
   &__further {
--- a/assets/css/pages/post.scss
+++ b/assets/css/pages/post.scss
@@ -172,6 +172,30 @@
     }
   }
 
+  // Last updated
+  &__last-updated {
+    padding: 0 calc(50% - #{$size__content-max-width} / 2);
+    background-color: $color-white-100;
+    @include full-width-background($color-white-100, true);
+
+    @include breakpoint('medium') {
+      &::before,
+      &::after {
+        display: none;
+      }
+    }
+
+    &-date {
+      @extend %font-ui-text-sm-oxygen;
+      color: $color-black-160;
+    }
+
+    &-separator {
+      margin-bottom: rem(48);
+      color: rgba(0, 0, 0, 0.2);
+    }
+  }
+
   // Further reading section
 
   &__further {
--- a/assets/css/abstracts/_placeholders.scss
+++ b/assets/css/abstracts/_placeholders.scss
@@ -187,12 +187,12 @@
 
 %page-content-bg {
   color: $color-black-190;
-  background: $color-page-background;
+  background: $color-page-content-bg;
 
   // Ensures that background extends to edges on mobile devices
   @include breakpoint($break: 'medium', $dir: 'max-width') {
     @include full-width-background(
-      $background: $color-page-background,
+      $background: $color-page-content-bg,
       $includeMargins: true
     );
 
--- a/assets/css/abstracts/_placeholders.scss
+++ b/assets/css/abstracts/_placeholders.scss
@@ -187,12 +187,12 @@
 
 %page-content-bg {
   color: $color-black-190;
-  background: $color-white-100;
+  background: $color-page-background;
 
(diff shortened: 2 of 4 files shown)


assets/css/pages/post.scss
```diff
@@ -172,6 +172,30 @@
     }
   }
 
+  // Last updated
+  &__last-updated {
+    padding: 0 calc(50% - #{$size__content-max-width} / 2);
+    background-color: $color-white-100;
+    @include full-width-background($color-white-100, true);
```
jnschrag, line 179: You shouldn't need to specify this, since it'll inherit the background of the parent post. This means you can also remove the pseudo elements too.

assets/css/pages/post.scss
```diff
@@ -172,6 +172,30 @@
...
+    }
+
+    &-date {
+      @extend %font-ui-text-sm-oxygen;
+      color: $color-black-160;
+    }
+
+    &-separator {
+      margin-bottom: rem(48);
+      color: rgba(0, 0, 0, 0.2);
```
jnschrag, line 195: When styling horizontal rules, you usually want to style the border property instead, so this would be `1px solid $color-black-120`.

assets/css/pages/post.scss
```diff
@@ -172,6 +172,30 @@
...
+
+    @include breakpoint('medium') {
+      &::before,
+      &::after {
+        display: none;
+      }
+    }
+
+    &-date {
+      @extend %font-ui-text-sm-oxygen;
```
jnschrag, line 189: I don't think this is the right placeholder. Figma says "post text/sm", which would be `%font-post-text-sm`.

jnschrag (changes requested):
This is looking good, José. A few things:

- Good catch on the background color! Yes, please update it. It doesn't look like that color (#f2f2f2) was included when we created the color variables, so you'll need to add it in `_variables.scss`. I would call it `$color-post-background`. Please use the RGB format to match the other colors.
- Can you test this on a different post, specifically one of the test posts? This one is a real post, and I don't want to have our test content included on it.
- The date should read "Updated [date]" (we're missing the "Updated" text).
- This CSS should be moved to `post-content.scss`, because this is a block in the content, and so technically could be used on other pages, not just posts. You could simplify the CSS class name to `last-updated__[block name]` too (eg. `.last-updated__date`).
- After looking at this in practice, I think the HTML structure is a bit off. Semantically, it doesn't make sense for the date to only be a paragraph. It functionally kind of serves the same role as a heading. I think the HTML should be updated to this:

```html
<div class="wp-block-lazyblock-last-updated-section">
  <hr class="last-updated__separator">
  <div class="last-updated">
    <h2 class="last-updated__date">Updated [Date]</h2>
    [Content ]
  </div>
</div>
```
- Because it's currently set to be a paragraph, the space between the date & the content is off. It should be `1rem`.

jdromero88:
Hey @jnschrag because the Glossary, About pages have a ```background color: #f7f7f7```
& after consulting with Tucker for consistency she advice to use the same color on the Analysis page, so I went ahead & update this file ```_placeholders.scss``` to

```
%page-content-bg {
  color: $color-black-190;
  // background: $color-white-100;
  background: $color-page-background;

  // Ensures that background extends to edges on mobile devices
  @include breakpoint($break: 'medium', $dir: 'max-width') {
    @include full-width-background(
      $background: $color-page-background,
      $includeMargins: true
    );

    &::before,
    &::after {
      z-index: -1;
    }
  }
}
```
I'm not sure if this the right approach but this is what makes sense in my brain. But let me know if is otherwise.

You can test the changes [http://localhost:3000/analysis/hello-world](http://localhost:3000/analysis/hello-world) and [http://localhost:3000/analysis/test-post](http://localhost:3000/analysis/test-post)

jnschrag (changes requested):
This looks great! Two small changes. One is a variable name I flagged, and the other is miscommunication on my part.

If you look at the HTML for the block in the browser:

Screen Shot 2021-05-27 at 11 29 34 AM

You can remove the div you added to the lazyblock with the class of `wp-block-lazyblock-last-updated-section`. There's a container div with that class automatically created by the plugin, so you don't need to add it.

jnschrag (approved):
Perfect, thanks so much, José!

PR #153, opened 2021-05-27
Part 1: Feat/linking to the glossary page 

jdromero88:
Hi @jnschrag here is the part 1 of the issue Glossary Terms in Analysis Posts # 150.
You can test it [here](http://localhost:3000/analysis/hello-world) and [here](http://localhost:3000/analysis/test-post)

### DIFF
--- a/store/analysis.js
+++ b/store/analysis.js
@@ -187,7 +187,10 @@ function contentFormatter(content) {
   let formattedContent = {}
   const content_parsed = parse(content)
   // console.log(content)
-  const linksToGlossaryFormatted = linkFormatter(content_parsed, content)
+  const linksToGlossaryFormatted = glossaryLinkFormatter(
+    content_parsed,
+    content
+  )
   // Grabs all the spans that have the id that we need
   let spansID = content_parsed.querySelectorAll('.easy-footnote-margin-adjust')
   // Grabs all the spans that contains the a tags
@@ -201,12 +204,14 @@ function contentFormatter(content) {
   return formattedContent
 }
 
-function linkFormatter(contentParsed, content) {
+function glossaryLinkFormatter(contentParsed, content) {
   // Grabs the links for the glossary
   let linksEl = contentParsed.querySelectorAll("[data-type='glossary']")
   if (!linksEl && !linksEl.length) return content
   linksEl.forEach((link) => {
-    const id = link._attrs.href.substring(38, link._attrs.href.length - 1)
+    const url = link._attrs.href.split('/')
+    const urlLength = url.length
+    let id = url[urlLength - 1] != '' ? url[urlLength - 1] : url[urlLength - 2]
     link.setAttribute('href', `../glossary#${id}`)
   })
   return contentParsed.toString()
--- a/store/analysis.js
+++ b/store/analysis.js
@@ -186,6 +186,7 @@ export const actions = {
 function contentFormatter(content) {
   let formattedContent = {}
   const content_parsed = parse(content)
+  // console.log(content)
   const linksToGlossaryFormatted = linkFormatter(content_parsed, content)
   // Grabs all the spans that have the id that we need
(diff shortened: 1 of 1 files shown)


jnschrag (changes requested):
Thanks, José! This is close, but the anchor links aren't working correctly. On the glossary page, the `slug` of the glossary term is used as the `id`. So for example, if I wanted to link to "Mean Orbital Speed" (which is the term the example in the Test Post is meant to go to), the url should be `/glossary#mean-oribital-speed`.

Screen Shot 2021-06-01 at 2 17 15 PM

Right now, the URL uses the Post ID as the anchor identifier, which means it doesn't go to the correct place. Ideally, we'd want to swap it out so it uses the slug instead, so we get a nice URL. But if that isn't possible, then we need to update the glossary page to use the Post ID as the anchor instead of the slug.

jdromero88:
Hi @jnschrag sorry, I totaly missed the id on the glossary. Now is updated & is working properly.
You can test it [here](http://localhost:3000/analysis/hello-world) and [here](http://localhost:3000/analysis/test-post)

store/analysis.js
```diff
@@ -186,19 +186,32 @@ export const actions = {
...
-    formattedContent.content = content
+    formattedContent.content = linksToGlossaryFormatted
   } else {
     formattedContent.content = formatContent(content_parsed, spansID)
     formattedContent.footnotes = formatFootnotes(spansAEl, spansID)
   }
   return formattedContent
 }
 
+function linkFormatter(contentParsed, content) {
```
jnschrag, line 204: Can we rename this to `glossaryLinkFormatter` so it's more specific on which kinds of links we're formatting?

store/analysis.js
```diff
@@ -186,19 +186,32 @@ export const actions = {
...
+  // Grabs the links for the glossary
+  let linksEl = contentParsed.querySelectorAll("[data-type='glossary']")
+  if (!linksEl && !linksEl.length) return content
+  linksEl.forEach((link) => {
+    const id = link._attrs.href.substring(38, link._attrs.href.length - 1)
```
jnschrag, line 209: While this works, it's also very dependent on the URL structure not changing, which makes me nervous because if we ever change the WPE URL, this would stop working and we'd have to figure out why. A more future-proof solution might be to instead `.split()` the URL on the `/` character to create an array, and then get either the last element in the array or the second to last element in the array depending on if it ends in a `/`. Something like this should get you started:

```
const url = link._attrs.href.split('/') // Creates an array.
const urlLength = url.length
let id = url[urlLength - 1] != '' ? url[urlLength - 1] : url[urlLength - 2] // This should check to see if the last element in the array is our glossary term (or an empty space/ending backslash) and return the correct array value accordingly.
```

This is still not entirely future proof, in that if we ever change the structure of the URL so the glossary term is not last, this would also break. But that's much less likely than us changing the URL & the amount of characters it has that the `substring` method relies on.

jnschrag (changes requested):
Thank you, @jdromero88! This is almost there!

store/analysis.js
```diff
@@ -201,12 +204,14 @@ function contentFormatter(content) {
...
-function linkFormatter(contentParsed, content) {
+function glossaryLinkFormatter(contentParsed, content) {
   // Grabs the links for the glossary
   let linksEl = contentParsed.querySelectorAll("[data-type='glossary']")
   if (!linksEl && !linksEl.length) return content
   linksEl.forEach((link) => {
-    const id = link._attrs.href.substring(38, link._attrs.href.length - 1)
+    const url = link._attrs.href.split('/')
+    const urlLength = url.length
+    let id = url[urlLength - 1] != '' ? url[urlLength - 1] : url[urlLength - 2]
```
jdromero88, line 214: hey @jnschrag thank you for opening my brain. this way of using the ternary operator just blows my mind!

jnschrag (approved):

PR #155, opened 2021-06-02
Added mounted hook to test the v-tippy

jdromero88:
Hi @jnschrag 
I can add it as a custom property to the options but, when I try to do the same to the element where we actually need it, it won't do it.
Screenshot at 2021-06-02 12-49-58
Not sure if I missed something in the [docs](https://vuejs.org/v2/api/#Instance-Properties).
Reading [this](https://vuejs.org/v2/api/#v-html) makes me think that we would need to take a different approach? Let me know what are your thoughts on this.

### DIFF
--- a/pages/analysis/_slug.vue
+++ b/pages/analysis/_slug.vue
@@ -234,6 +234,33 @@ export default {
       users: (state) => state.analysis.users
     })
   },
+  vueTippy: 'v-tippy',
+  mounted: function() {
+    this.$nextTick(function() {
+      // Code that will run only after the
+      // entire view has been rendered
+      // I can add it as a custom property to the options but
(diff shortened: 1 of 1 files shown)


jnschrag:
Thank you for digging into this, @jdromero88! I think you're right - I don't think this is going to be possible the way that we're currently rendering content, which is unfortunate but not that surprising. I know that there are ways to be able to use Vue components within content, such as the [Nuxt Content Module](https://content.nuxtjs.org/), but it would require an overhaul of how the content system works and I don't know that it's worth it for this piece of functionality at this point. I'm going to close this PR for now, and update the issue accordingly. I'll review #153 so we at least get the glossary links working so it redirects to the correct page on the site.

Thanks so much for doing the research on this. I really appreciate it!

Issue #156, opened 2021-06-07
Actions Fix AstriaGraph error status code


HP4k1h5:
In some cases the curl request fails with 502 e.g.

```
Run curl -s -X POST "${TOKEN}"
<!DOCTYPE html>
<html>
<head>
    <title>Error Loading Site | 502 Bad Gateway</title>
</head>
```

`curl` typically exits with a 0 exit code in such cases but it would be helpful to know when this script is failing. 

this should be possible with a combination of `curl` flags

PR #157, opened 2021-06-07
fix: #156 github action astria-graph error status

HP4k1h5:
this fix hopefully addresses the zero exit status of curl when it
returns a non 2XX response status code. it may not fully address this
issue but is hard to test without running on the github actions server

 Changes to be committed:
	modified:   .github/workflows/astria-graph.yml

### DIFF
--- a/.github/workflows/astria-graph.yml
+++ b/.github/workflows/astria-graph.yml
@@ -12,6 +12,6 @@ jobs:
     
     steps:
     - name: Trigger our build webhook on Netlify
-      run: curl -s -X POST "${TOKEN}"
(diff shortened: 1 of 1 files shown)


jnschrag (approved):
Thanks, @HP4k1h5! I admittedly don't really understand what these flags do. I did a quick look up & it technically seems right so I'll mark this as "approved", but could you elaborate a bit on how you think it'll help this particular error that we're getting? Thank you!

jnschrag:
Got it. Thanks so much for that explanation. I'll merge it in and we can see if it results in a change!

PR #159, opened 2021-06-08
Update min distance to km

jdromero88:
Hi @jnschrag 
if I update the line 239 to ```min_distance_km```
Screenshot at 2021-06-08 17-28-19
the table won't be populated & so far I couldn't figured it out yet.
Screenshot at 2021-06-08 17-32-06

### DIFF
--- a/pages/key-events.vue
+++ b/pages/key-events.vue
@@ -236,7 +236,7 @@ export default {
             type: 'desc'
           },
           {
-            field: 'min_distance', // if change to min_distance_km the table is empty
+            field: 'min_distance',
             type: 'asc'
           }
         ],
--- a/pages/key-events.vue
+++ b/pages/key-events.vue
@@ -150,8 +150,8 @@
                 </div>
               </div>
             </template>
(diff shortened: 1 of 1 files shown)


jnschrag:
> if I update the line 239 to `min_distance_km`
> Screenshot at 2021-06-08 17-28-19

Thanks for digging into this, @jdromero88. This isn't your fault - I gave you incomplete instructions/this required some background knowledge you had no way of knowing. The `serverParams` are what is passed to the WordPress server so the API call returns the correct dataset. In this case, you're modifying the `sort` parameter on the database query. The reason why the table is blank when you change it to `min_distance_km` on this line is because the `min_distance_km` field is derived from the `min_distance` field, but does not actually exist in the database. So the end result is that this API call is trying to query on a table column that doesn't exist, and so is returning no results.

The solution is to leave the `serverParams` as they are, and then modify the areas where the distance is being displayed so it uses `min_distance_km`.

jdromero88:
Hi @jnschrag thank you for the clarification!!

jnschrag (approved):
Thanks so much, José!

PR #161, opened 2021-06-14
Draft: Feature/add vue highcharts hist longitude

### LINKED ISSUE #50: Component: Magic Chart
This modal opens up the "Magic Chart" component that shows the historical longitudes of the two selected objects. It uses the modal component (#48)

- [x] Title: `Compare Objects`
- [x] Disclaimer
- [x] Data Sources
  - [x] Satellite Name
  - [x] Source Name
  - [x] Source timestamp of most recent position
- [x] Print button
  - [x] Print styles for chart
- [x] Download
  - [x] CSV
  - [x] JPG
  - [x] SVG

## Chart
- [x] Title: `Historical Longitudes`
- [x] Y Axis
  - [x] Tickmarks: 180, 0, 180
- [x] X axis (time based)
- [x] Lines per satellite
  - [x] Dashed lines for predictions (pulled from close approaches data)
- [x] Current date indicator
- [x] Legend
- [x] Drag functionality to change zoom level of chart
  - See [Baby Names](https://nbremer.github.io/babynames/) chart for example
  - See [Focus + Context](https://observablehq.com/@d3/focus-context) code example

HP4k1h5:
close #50


HP4k1h5:
Merge depends on https://github.com/CSIS-iLab/satellite-dashboard-backend/pull/20

after merging backend 20, remove sorts and TODOs from this pr

jnschrag (changes requested):
Thanks so much for working on this, Bob! I think this is really close to being finished and I know the program is going to be ecstatic to see it finished. I've got a lot of notes that mostly have to do with the appearance to get the charts to match up more closely to what are in the mockups, but functionally I think both charts are great. I know this is a lot, so if you want me to jump in and help with this at all, please let me know!

# Historical Longitudes (Small Chart)
- [x] Can we remove the export button? We don't want that to be an option on this source
- [ ] Let's remove the Highcharts credits. Their license allows for that, and if needed we can credit them on the About page or in the site footer.
- [ ] The background of the chart should be transparent
- [ ] Can you check the font color & size of the axis labels? Those look off. They should be `$color-white-160` and use the `%font-ui-text-xsm-oxygen` placeholder.
- [ ] Likewise, the chart title also seems off. It should match the styles for the "Close Approaches" heading below it.

# Magic Chart
- [ ] Is there a way to make the initial modal/loading state more closely matching the size of the eventual modal? If not, then can we at least increase the size a bit so the title of the modal isn't running into the close icon?
- [ ] The title of the modal typography doesn't match the mockups anymore. I think because the `<h3>` isn't inheriting the font styles from the parent `.modal__header` class.
- [ ] Same note for the title of the chart, the typography looks a little off. 
- [ ] Can you look at the typography for the axis labels? The Y axis label is too large and not the correct color, and the smaller subtick labels aren't the same color as the mockups either.
- [ ] We do want to show the Y axis grid lines, but not the X axis.
- [ ] On the X axis, can we list it as Month Day when zoomed in at that level?
- [ ] The typography for the legend (the satellite names) is off. We're also missing the satellite ID numbers.
- [ ] Question: Both satellite's "overall" views are represented as blue in the overview/full lifetime chart. Is there a way to get them to match their respective color? If not, then can we make them gray lines like the mockup?
- [ ] The data sources section typography & spacing doesn't match the mockups.
- [ ] Reminder that we'll need to remove the export button in the top left & move that functionality to the individual buttons at the bottom of the modal


# Other
- [ ] The "Compare Objects" button in the close approach list should have these classes on it `btn btn--outlined` and then be aligned so it's in the center.

# Questions for Thomas
- Do we want to have the tooltip enabled on the small chart. If he does, then we should format the date so it matches the style used elsewhere on the site (eg. Oct 9, 2020 at 00:00:00 AM) and we should only show up to 4 decimal places for the degrees.
  - We should ask him to review both chart's tooltips so he can let us know how he wants that information formatted.
- We should also ask him if he wants us to show the E/W degrees on the Y axis and in the tooltip on both charts
- We should ask Thomas about the default view of the chart - zoomed in or all the way out/ask if he wants to keep the "Zoom YTD, 1y, All" buttons
  - [ ] If he does, then let's format those buttons like the "Download buttons & label at the bottom"
  - [ ] If he wants to keep the date selection, we should format it like other date selections on the site for consistency (click the date on the timeline to see an example)

jnschrag (changes requested):
Thanks so much for taking another look at this! Overall it's looking a lot better, but there's still a lot of mismatches between the typography & colors outlined in Figma vs what's on the site. Let me know if you want to discuss and I can show you where I'm getting these values.

# Historical Longitudes Chart
- [ ] Is there a way to eliminate the space on the left side of the chart, so that the axis labels & chart title line up with the close approaches text below it? I know Highcharts adds some padding by default. If that isn't possible, than an alternative might be to offset that padding added by Highcharts with a negative margin on just that section of the panel so it looks lined up.

Screen Shot 2021-07-08 at 10 46 12 AM

- [ ] The title text "Historical Longitudes" is too big (should use the `%%font-ui-heading-section-md` placeholder with `font-weight: bold; letter-spacing: 0.025em` applied) and isn't the correct color (should be `$color-white-180`). I know that working with the titles within Highchart itself can be tricky, so if that's the case, then maybe it's better to render the title in HTML and not a part of the chart if that's an option.
- [ ] The Y axis & endpoint labels aren't the correct size or color on this chart either. Note that these axis labels are different from the ones on the Magic Chart - that is intentional. The orange color on the endpoint labels does match!
- I'm not seeing the Y axis grid lines on this chart. Any idea why that would be?

Screen Shot 2021-07-08 at 10 55 13 AM

- Is there a way to not show the upper/lower most Y axis values here? In the example I'm looking at, the 80/100 marks should be enough since the values fall below that. This isn't a huge deal, but would help condense the chart a little bit more if we could remove any extra labels.

# Magic Chart
- [x] Can you check the typography on the "Historical Longitudes" title? It looks like it's using the Oxygen font there instead of the Good Pro like it's meant to.
- [x] The Y axis title and Y/X axis labels are all off in terms of their size & color.
- [x] The satellite names in the legend should be `font-weight: normal; font-size: 14px`
- [x] Thanks for adding the satellite ID numbers. The color is correct, but the sizing is off on those.
- [x] The footer section where it says "Data Sources" is in the wrong space. It should be a part of the main chart area (the darker bg/modal footer is for the print & download buttons). Double check the typography here too - everything is just a little too big.
- I actually don't see the Y axis grid lines here. The only line I see is a very bright line for the X axis. Very odd :/

Screen Shot 2021-07-08 at 11 07 23 AM

# Other
- [x] The "Compare Objects" button isn't centered in the space. It shouldn't need to be inside of a div that has the `.close-approaches__item` class (that class has other CSS applied to the main container for each list item). Instead, let's give it's own class and add this to it to center the button:
```
grid-column: 1 / -1;
justify-self: center;
```

Screen Shot 2021-07-08 at 10 59 51 AM

assets/css/components/close-approaches.scss
```diff
@@ -74,3 +78,8 @@
     font-style: italic;
   }
 }
+
+.compare-objects {
```
jnschrag, line 82: Can you update this class name so it follows the naming pattern of the rest of the component? Something like `.close-approaches__compare-objects`?

assets/css/components/highcharts-theme.scss
```diff
@@ -0,0 +1,97 @@
+@import '~/node_modules/highcharts/css/highcharts';
+
+.highcharts-root {
+  font-family: $font__good;
```
jnschrag, line 4: This should be Oxygen, since that's the primary font used for most of the labels and such. It's just the headings that use the Good Pro font.

assets/css/components/highcharts-theme.scss
```diff
@@ -0,0 +1,97 @@
...
+}
+
+.highcharts-title,
+%font-post-heading-h3 %font-ui-heading-section-lg {
```
jnschrag, line 8: This title doesn't look right, but I think that'll be corrected when the placeholders application is fixed.

jnschrag (changes requested):
This is looking really good! I've left a few comments in specific spots, but this is a lot closer to the specs which is great!

# General
- I think the CSS isn't being scoped properly - some of the styles from the details chart are being applied to the Magic Chart and vice versa. It's probably better to scope those, either when importing the CSS files to the component or by nesting the highchart class under the unique selectors for each chart (eg `.details-chart .highcharts-root` vs `.magic-chart .highcharts-root`).

# Details Chart
- [x] If we're removing the tooltip from this chart, can we remove the blue circle that appears on hover?
- [ ] When the Y axis label has 4 digits, it's cut off just a little bit. Is it possible to move the Y axis labels over just a few pixels?

Screen Shot 2021-07-09 at 11 05 37 AM

# Magic Chart
- [ ] The "Data Sources" section typography isn't correct. The satellite names should use `%font-ui-text-base-oxygen` with `font-weight: bold` applied. Right now they're using the Good Pro font and a different font size & weight. The "Data Sources" label itself also doesn't match the specs.
  - [ ] Can we match the spacing between the dividing line and "Data Sources"? It should have `rem(12)` of space between it and the title. The sides of that section should be flush with the side - they're currently indented a little bit because of the `padding: 1%` applied on the data sources container.
- [ ] Can you add the below CSS to the print button so the icon & text line up with each other?
```
display: flex;
align-items: center;
gap: rem(4px);
```
- [ ] Take a look at the padding around the data sources & the download buttons. The edges of the print & SVG buttons should line up with the ends of the line above the data sources section.
- [ ] Can we increase the space between the CSV, JPG, and SVG buttons?
- The export functionality works great, but the colors are off. Presumably because the chart is styled to work on a dark background, not a light one. Can you look and see if there's an option to apply export styles to a chart? If so, we'll probably want to use CSS variables to control these so we can switch them from light to dark based on if we're viewing the chart in the browser vs exporting it (this can be a separate issue to not hold this up).
- Have you noticed any issues with the drag functionality? In the below screenshot, the area that I've dragged does not align with the dates shown in the date selector above the chart, or the data on the chart.

Screen Shot 2021-07-09 at 11 24 20 AM

assets/css/components/details-chart.scss
```diff
@@ -0,0 +1,55 @@
+.details-chart {
+  max-width: 100%;
+  max-height: 100%;
+  margin-left: -10;
```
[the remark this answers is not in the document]
HP4k1h5: thanks, bad copy job from highcharts styles

HP4k1h5:
Thanks for your notes. ive made these changes as best i can. One issue i keep having is that when my styles match the figma spec, the output ends up looking quite different from the figma mockups. Prior to these recent changes i was going more by eye, than by spec, but i'm defaulting to the spec now. Apologies for the multiple rounds of corrections.

Getting the export to resemble the charts better is going to take more work. After some recent changes, i'm now seeing a dark background for export (see attached), but getting the just right will probably be a larger effort. You can add styles to the export functionality, but it's going to take some work to know which styles are automatically included and which need to be forecfed. 

I have not noticed the issue you flagged with mismatched master-detail timeframes. 
Screen Shot 2021-07-09 at 2 22 00 PM
Mine do not seem to have this issue. I might recommend, `rm -rf node_modules package-lock.json && npm i` as well as full server restart and browser refresh. There can be deps discrepancies when loading a branch with package dep diffs.

historical-longitudes-9

assets/css/components/highcharts-theme.scss
```diff
@@ -0,0 +1,98 @@
+@import '~/node_modules/highcharts/css/highcharts';
+
+.highcharts-root {
+  font-family: $font__oxygen;
+}
+
+.highcharts-title {
+  @extend %font-post-heading-h3;
```
jnschrag, line 8: There should only be one placeholder here, the `%font-ui-heading-section-lg`. If there's anything else that needs to be styled differently from the placeholder, like font weight, that can be declared on this selector too. But typically we shouldn't have multiple placeholders extended like this.

jnschrag (changes requested):
Thanks for working on this. Besides the placeholder stuff we discussed over slack, I only have a few other comments. Thanks for the tip on the packages. I reinstalled everything and the drag functionality now works. When I reset it, the sass division error we talked about for PONI popped up, so I made this change to line 15 of `css/abstracts/_functions.scss` if you'd like to commit it (`$rem-size: math.div($size, $browser-context);`).

- [x] I think we need to scope the CSS for the two charts - they're conflicting with each other still.
- [x] I see the "Today" text on the close approach for NSS 9 & YAMAL 300k predicted for July 13th. Can you add some padding on the right side so the text isn't cut off, and update the color to something other than black?

Screen Shot 2021-07-12 at 11 08 52 AM

- [x] The left/right spacing around the data sources area doesn't match up with the footer of the modal. They should vertically line up with each other.

Screen Shot 2021-07-12 at 11 09 00 AM

assets/css/components/highcharts-theme.scss
```diff
@@ -0,0 +1,118 @@
...
+
+.highcharts-axis-line {
+  stroke-width: 0;
+}
+
+.highcharts-grid-line {
+  stroke: $color-white-110;
+}
+
+.highcharts-axis-title {
```
jnschrag, line 32: Can you double check this one? The separate font family & font weight properties probably don't need to be here if we're using the correct placeholder.

assets/css/components/highcharts-theme.scss
```diff
@@ -0,0 +1,118 @@
...
+  @extend %font-ui-text-lg-oxygen;
+  font-family: $font__good;
+  font-weight: 400;
+  opacity: 0.8;
+  fill: $color-white-180;
+}
+
+.highcharts-legend text {
+  font-family: $font__oxygen;
```
jnschrag, line 41: Same here - there's probably a placeholder to go with this one.

assets/css/components/magic-chart.scss
```diff
@@ -0,0 +1,89 @@
+@import '~/assets/css/components/highcharts-theme';
+
+.magic-chart {
+  min-width: 45vw;
+  min-height: 45vh;
+}
+
+.highcharts-axis-title {
+  @extend %font-ui-text-lg-oxygen;
+  font-family: $font__good;
```
jnschrag, line 10: Same note as above - this probably needs to use a different placeholder so the correct font attributes are all there.

assets/css/components/magic-chart.scss
```diff
@@ -0,0 +1,89 @@
...
+
+.highcharts-axis-labels text {
+  @extend %font-ui-text-sm-oxygen;
+  opacity: 0.4;
+  fill: $color-white-180;
+}
+
+.legend-id {
+  font-weight: 100;
+  font-size: 10px;
```
jnschrag, line 24:
```suggestion
@extend %font-ui-text-xsm-oxygen;
```

assets/css/components/magic-chart.scss
```diff
@@ -0,0 +1,89 @@
...
+  @extend %font-ui-text-xsm-oxygen;
+  fill: #abadaf;
+}
+
+.data-sources {
+  padding: rem(12) 0 0 0;
+  border-top: 1px solid $color-white-115;
+}
+
+.data-source-title {
```
jnschrag, line 38: This should use `%font-ui-text-lg-good`.

jnschrag (changes requested):
Just a couple more typography things and then I think this is good to go.

As a FYI, if you're not sure which placeholder to use, you can usually derive the name based on the comment above the font properties in the "Inspect" tab in Figma.

Screen Shot 2021-07-12 at 4 10 30 PM

This would translate to `%font-ui-text-lg-good`. Sometimes there is variations, but it should get you close enough that you can look it up in `_placeholders.scss`.

Also, I think I wasn't clear about the data sources spacing thing. The spacing in the footer of the modal was correct - the "data sources" section needed to match it.

Screen Shot 2021-07-12 at 4 05 18 PM

It should be `rem(32)` on each side of both the data sources & the modal footer. In this case that value wasn't marked, but you can find it (or other spacing in Figma) by selecting a layer and hovering over other elements. It should automatically draw lines and tell you the distance between them. Sometimes it isn't a perfect multiple of 8 - if that happens, round it to the closest multiple of 8 (eg. if this had shown me 30px, I would have rounded it to 32px).

Screen Shot 2021-07-12 at 4 16 19 PM

jnschrag (approved):

PR #162, opened 2021-06-25
Bug - Filter Results: sort

jdromero88:


components/visualizer/FilterResults.vue
```diff
@@ -131,17 +129,28 @@ export default {
...
+            return x[0].id < y[0].id ? -1 : 1
+          }
         },
-        { label: 'CO.', field: 'country', sortFn: this.sortFn },
         { label: '', field: 'actions' }
       ],
       sortOptions: [
-        { value: 'Name', label: 'Name' },
-        { value: 'country', label: 'Country' }
+        { value: 'Name A-Z', dir: 'asc', label: 'Name A-Z' },
```
jnschrag, line 144: Why does the `value` include the `A-Z` part? Shouldn't that be implied by the `dir` property?

jnschrag (changes requested):
This is looking really good and works well. Thank you both for looking into it! I had one comment about something in the code I was confused about. The only change is a styling issue. The label should all be on one line, which can be done by setting `white-space: nowrap` on the `.vs__dropdown-toggle` class.

Screen Shot 2021-06-30 at 11 59 21 AM

I'll also want to talk to Tucker about the actual labeling for these options, but that's easily fixed!

components/visualizer/FilterResults.vue
HP4k1h5: i wanted the `value` to be unique because i think the table depends on it. i added the dir for simplicity sake buts its not a native vue-table field

jnschrag (approved):

PR #164, opened 2021-07-02
Track orbital elements on date slider change

samadd:
This allows components to be notified on date slider change, and update the selected orbital element value accordingly. Incomplete element sets are padded in order to match the number of days in the selected timescale.  This was previously done in the CesiumViewer component.


jnschrag (approved):
Thanks, @samadd! This looks great. Bob and I both took a look and it's working which is so, so exciting!

Can you remove the `console.log` statements you added in here to test it? Then I can approve it and merge it in.

There were two things that Bob & I noticed that I think might be separate bugs (but maybe not):
1. For me, when I went back in time to see previous dates, the satellites all disappeared and never reappeared. But if I focused on a satellite first for the current day, then went back in time so the details pane was still open, I could see the values update when I pressed play.
2. For Bob, the satellites took a second to reappear but did eventually show up, but the focus on the selected satellite disappeared (the orbit line disappeared). I'm less fussed about this one, but if that's a straight forward fix to maintain the focus state of the satellite, that would be awesome.

Let me know what you think and if you want to look at these as separate issues or as a part of this one.

samadd:
That's the console.log removed! Not that I rely on console.log for all my debugging... :)

As for those two issues:
1. When I try to replicate, I see that the satellites disappear but I'm not convinced that it's anything to do with this, because the API response itself seems to be broken. When I load data for today, getting elements from the 4th, that loads fine. But if I switch back to, say, 27th July, the response cuts off and leaves invalid JSON, so the cesium component throws an error because the data arrives as a string, since presumably Axios has tried to JSON.parse it, failed, and just left it as a string. I've also been seeing some occasional CORs errors too, but not every time. Sounds like something's up on the back end?
2. Keeping satellite focus and orbital path on date change is a good shout  - maybe worth doing as a separate task?

I've also been lookng at the focus list panel requirement you mentioned - where it indicates whether the set of elements is complete, patchy or non-existent for the date range. I'm on the verge of having something working for that, but wanted to check whether you had an idea of what those indications should look like, ie just some text, or something else? And if I should continue that in a new branch...?

jnschrag:
Thanks, Sam! You've seen my code, you know I have no issues with good old `console.log` for debugging. :)

1. Thanks for flagging. @HP4k1h5 has been doing some work on the backend that I think might fix this. Let's put a pin in it for now and if it's still becoming an issue, maybe the 3 of us could chat and we can figure out a solution. Now that we've got all the historical data in, we're working with just over 4 million rows of orbital paths, so I'm not super surprised that we're running into issues getting it seamlessly delivered to the front of the app.
2. Yeah, let's make this a separate task!

For the focus list, let's go ahead and make that a separate branch. I'll merge this one in now just so it's done. For reference, that note is just text that appears on the list item in question. You can check it out in [Figma](https://www.figma.com/file/XzGzlXpNmeB3PzTNfzk4ro/Satellite-Dashboard%2C-handoff?node-id=40%3A24851) or in the screenshot below. 

Screen Shot 2021-07-07 at 11 02 14 AM

If you want to implement the functionality bits of it, I'm more than happy to style the text later.

samadd:
Nothing wrong with a bit of console.log :)

1 - Hopefully that issue will disappear, then!
2 - I'll make that fix in a new branch

Thanks for the focus list screenshot - I don't seem to have access to the
Figma. My only question is around the text - when we have, say, 3 days of
matching orbital element data over a week, I'm showing a disclaimer
whenever we're on a date that doesn't have matching data and is thus
derived from the closes available date. That doesn't seem to be captured by
"this object is not yet in orbit". Do we have alternative disclaimer text?

jnschrag:
> Thanks for the focus list screenshot - I don't seem to have access to the
Figma. My only question is around the text - when we have, say, 3 days of
matching orbital element data over a week, I'm showing a disclaimer
whenever we're on a date that doesn't have matching data and is thus
derived from the closes available date. That doesn't seem to be captured by
"this object is not yet in orbit". Do we have alternative disclaimer text?

Good question. I actually think the use case is a bit different (so maybe it's not related). This disclaimer should show if we don't have _any_ data for a given satellite in the selected time frame. If we have 3 days out of 7, and are padding out the missing days based on the data from those 3, then I don't think showing the disclaimer is necessary. This disclaimer would be if we have absolutely no data for this satellite for the given time frame. Does that make sense?

samadd:
Yes, makes perfect sense! And means that I can simplify the code I've
already written :)

PR #165, opened 2021-07-06
Update country field information

jdromero88:
Hi @jnschrag
Not sure if here `CloseApproachesList.vue` we should show the country but for now I added it. 
Screenshot at 2021-07-06 13-12-05

### DIFF
--- a/store/satellites.js
+++ b/store/satellites.js
@@ -211,7 +211,8 @@ export const actions = {
           let countryOfJurisdiction = formatCountries(sat.countryOfJurisdiction)
           let countryOfJurisdictionIds = countryOfJurisdiction.map((d) => d.id)
 
-          let countryOfLaunch = formatCountries(sat.countryOfLaunch)
+          // let countryOfLaunch = formatCountries(sat.countryOfLaunch)
+          let countryOfLaunch = formatCountries(sat)
 
           // Store the countryOfJurisdiction so we can filter on it later.
           countryOfJurisdiction.forEach((country) => {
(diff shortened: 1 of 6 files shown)


store/satellites.js
```diff
@@ -285,10 +284,12 @@ function formatCountries(value) {
   if (value === undefined) {
     value = ''
   }
+  let countryID = value != undefined ? value : 'TBD'
```
jnschrag, line 287: Why was this added?

jnschrag (changes requested):
Thanks for looking at this, @jdromero88! This all looks really good. I had one question about a particular line that I left. Besides that, can you go ahead and just remove any references to `countryOfJurisdiction` instead of commenting it out? We're going to full on remove that field from the site, so let's just get rid of it.

store/satellites.js
jdromero88: Oh I forgot to remove this, but for testing purposes I was using it in a different way. On line 198 I was passing the whole object sat:
```
let countryOfLaunch = formatCountries(sat)
```

jnschrag (changes requested):
Hey José, thanks for this. I think I miscommunicated what I wanted. We're _replacing_ `countryOfJurisdiction` with `countryOfLaunch`. We don't just want to eliminate `countryOfJurisdiction`, we need to replace it so any functions/data/etc. that uses that field now uses `countryOfLaunch`. So in a file like `store/satellite.js`, we actually need to keep all of those functions that were removed, but replace it so it uses `countryOfLaunch`. Does that make sense?

jnschrag (approved):
Note that we still need to update the countries list to work with multiple countries.

PR #166, opened 2021-07-08
Add reset button to visualizer

jnschrag:
This adds a reset button to the visualizer, which purpose is to:
* Clear Filtered List (0 filters)
* Clear Focus List (0 focused satellites)
* Reset Visible satellites (should show all satellites)
* Reset date on timeline (go to current date)
* Reset speed scale (go to default speed)
* Reset time scale (go to default scale)
* Reset cesium camera view to default
* Reset cesium focused objects to none

This function does this by resetting parts of the satellite & filters states to their defaults. I followed the pattern outlined in this [GitHub comment](https://github.com/vuejs/vuex/issues/1118#issuecomment-356286218) on the issue. I left the parts of state that we do not want to replace, such as the satellite IDs, time scales, etc. in the main `state` declaration.

This functionally works really well, but I do think there's room for improvement. If you watch the console when this runs, you'll see that the `processNewData` function in `CesiumViewer.vue` is running twice. Once after you click the reset button because the visible satellites value is being reset, and again after `getOrbits()` is called (which pings the API to get the correct orbital data for the default view we're resetting too).

Screen Shot 2021-07-08 at 4 20 30 PM

My question is if it's possible to not have Cesium run that first `processNewData` since there's really no point until the new orbital data has been added. @samadd if you have any thoughts here since you implemented this, I'd be grateful! If not, then it's not the end of the world, but would be nice to try and optimize if we could.


samadd:
Would it not be the case that on the first processNewData call, the list is empty anyway since you've reset that state? Then it's not until you call getOrbits that there's any data to actually work on. So if that's the case, my thought would be that the call to processNewData is doing so little that there would be no point in trying to work around the data binding to prevent it from running, unless I've missed something?

HP4k1h5:
agree, the list is empty, but cesium does send out a few dozen requests. a check like 
```javascript
      if (!Object.keys(this.satelliteOrbits).length) return
``` 
at the top of `processNewData()` prevents this

jnschrag:
Thanks, both! I added an early return statement just in case. It might be a "micro" optimization, but this app could use all the help it can get at this point. :)

HP4k1h5 (approved):
nice work!

samadd:
Sounds good to me :)

PR #168, opened 2021-07-09
edit: move fetch to id watcher

### LINKED ISSUE #167: Bug: close approaches in details list gets out of sync
Description:
Clicking on a orbital object point brings up the "Details Panel". In the "Key Events" tab, there is a list of "Close Approaches" which should be a list of single objects that approach the clicked object. 

On first satellite click, the list has the correct data, however if you click on new objects, the list starts to exhibit incorrect behavior. You start to see satellite pairs, instead of singles, and then on further clicks the list doesn't refresh at all, and you have stale data.

Presumably the close approaches are being appended to, instead of replaced. 

Possible fix:
Change the mutation to replace close approach objects instead of appending. Use caution because this component is reused in at least one other place, in the "Latest Updates" Popup component.

HP4k1h5:
need to (re)trigger the fetch call every time sat id changes

Close #167

### DIFF
--- a/components/visualizer/details-panel/Events.vue
+++ b/components/visualizer/details-panel/Events.vue
@@ -102,12 +102,6 @@ export default {
       default: ''
     }
   },
-  async fetch() {
-    const events = await this.$axios.$get(
-      `/wp-json/satdash/v1/close_approaches/${this.id}`
-    )
-    this.events = events
-  },
   data() {
     return {
       activeTab: 'events',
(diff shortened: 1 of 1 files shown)


jnschrag (approved):
Thanks for looking at this! There is a split second where it still appends to the existing list, but it does eventually clear out and replace the whole list with only one. So I think that's okay. I'm gonna go ahead and approve this and if we need to dig back into it later, we can.

PR #169, opened 2021-07-09
Feat/share view button

jnschrag:
This adds a "copy link" button to the focus list. The purpose is to give the user an easy way to share a link to any pinned satellites at a given day/time. Note that I did make a decision to change the language than what is shown in the mockups - that was intentional to be more explicit about what this button functionally does.

The URL format is this: `?satids=11570,21052,16667&date=2021-02-28&time=86400&timescale=week`

This implementation doesn't currently add a `time` parameter because I couldn't figure out how to get the specific time out of the timeline. @samadd, do you have any ideas here? I feel like I'm overlooking something super obvious. If there isn't, then I don't think it's the worst thing in the world that this always defaults to 12:00am on the given date, and the `time` parameter can be used internally for linking to specific events where the time is known.


HP4k1h5 (commented):
This looks great. The copy functionality works.

It might just be me, but the "Copied!" tippy tooltip is "stuck", and appears almost all the time. I can make it go away and then behave properly, but typically it's just stuck there. 
Screen Shot 2021-07-12 at 9 56 33 AM

HP4k1h5 (approved):

PR #177, opened 2021-07-13
Show multiple countries in results list

### LINKED ISSUE #171: Bug: Display multiple countries
Need to make sure that we can easily display multiple countries in the below areas:
- [x] Filter list results
- [x] Details Pane
- [x] Focus List results

jnschrag:
Closes #171 by making sure that in all places where a 3 letter ISO code is shown for a satellite's country, it will show multiple countries if a satellite is associated with more than 1 country. Also does a few design tweaks to fix alignments.

Screen Shot 2021-07-13 at 5 03 46 PM

### DIFF
--- a/assets/css/components/close-approaches.scss
+++ b/assets/css/components/close-approaches.scss
@@ -69,6 +69,8 @@
     .sat__actions {
       grid-row: 1 / 3;
       justify-content: flex-end;
+      align-self: center;
+      text-align: right;
     }
   }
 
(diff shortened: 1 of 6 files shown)


HP4k1h5 (changes requested):
im seeing some bad line breaks for entries with multiple countries

Screen Shot 2021-07-14 at 9 36 52 AM

jnschrag:
Good catch! I've made some modifications to this area that should fix it for 99% of cases. There still might be some combinations that result in some broken line funkiness, but I think we did our best to account for them all.

I also noticed that this whole table was actually overflowing the parent container, so there was a horizontal scroll. I've made some modifications to the table padding to remove that.

HP4k1h5 (approved):

PR #178, opened 2021-07-13
Show disclaimer on focus list of no orbits are available

samadd:
This exposes the orbit details to the Focus List component. If no orbits matching the satellite ID are there, or the list is empty (not sure if the API would ever supply an empty set, but in case it does) then the disclaimer is shown.


jnschrag (approved):
Looks great, thanks Sam! I made some style tweaks, but otherwise this was exactly what we were looking for. :)

PR #179, opened 2021-07-14
Bug/fix details chart

HP4k1h5:
- longitude formatting: E/W vs +/-
- details simplifications
- query improvement

### DIFF
--- a/components/visualizer/MagicChart.vue
+++ b/components/visualizer/MagicChart.vue
@@ -90,7 +90,10 @@ export default {
         yAxis: {
           opposite: false,
           title: { text: 'Longitude', margin: 30 },
-          labels: { format: '{value}°' }
+          labels: {
+            formatter: (label) =>
+              `${Math.abs(label.value)}°${label.value < 0 ? 'W' : 'E'}`
+          }
         },
         tooltip: {
           formatter: function() {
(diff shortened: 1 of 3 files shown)


components/visualizer/MagicChart.vue
```diff
@@ -99,7 +102,9 @@ export default {
                 return `${s}<span class="tooltip-line-color" style="color: ${
                   i ? 'rgba(254, 116, 20,1)' : 'rgba(82, 175, 225, 1)'
                 }">–</span><span class="tooltip-y">
-                  ${point.y.toFixed(2)}°</span>`
+                  ${Math.abs(point.y).toFixed(2)}°${
```
jnschrag, line 105: Do you need to do this here? I know for the satellite orbital data, I had included the longitude data with the formatted label too (so that was all done server side). I don't remember what the longitude endpoint returns, but if it returns both the raw value and the label, then this should probably use the label. If it doesn't, ignore me.

jnschrag (approved):
Looks good to me. I had one question, but it's dependent on how you set up the API endpoint.

Is it worth abstracting out the label formatting to its own function since it's reused so many times on the chart?

If you're happy with this, feel free to merge it in.

components/visualizer/MagicChart.vue
HP4k1h5: great point, the `convert_longitude_format` does bring back both, but i only included the `degrees` field in the return vals. i can update the backend, but i'm not sure that 2 points of floating precision is enough for some of these charts

jnschrag: Good point. I think in my head the idea was you would use the super precise data to plot, and the formatted label to label...but at that point, it's probably just better to format the data on the front end instead of the back so you're not returning twice as much info. 100% your call on how you want to handle, I can see pros/cons to both of them. :)

HP4k1h5: also just thinking about this more, there isn't really a great way to map the labelled vals back to their floating point values once you pass this callback to highcharts. these callbacks have no context, no `this` and it would be almost impossible to align these vals with a separate array of labelled points

jnschrag: Sounds good. Then let's leave this as is!

Issue #126, opened 2021-02-19
Outstanding Items & Time Estimates


jnschrag:
This is an attempt to create a single, master list of all the remaining items necessary for Phase 1 of this project. Individual issues will be left open & updated as items are completed, but this list exists to provide an overview at a glance, including estimates on how long each task will take. The tasks have been roughly grouped by page or part of the process.

**Key**
🐛 Bug
⚡ Optimization

## Global
- [ ] Glossary Term Tooltip
  - [ ] Create tooltip component
  - [ ] Incorporate into Visualizer
  - [ ] Incorporate into Analysis
    - [ ] Create WP block/assess how editors can easily link to glossary terms when authoring posts

## About
- [ ] Creative Commons logo

PR #180, opened 2021-07-14
Fix details pane crash when no orbital data

### LINKED ISSUE #44: Component: Object Details
The Object Details panel opens when a user clicks on a satellite in the visual or on the focus list. Clicking on another satellite updates the information in the panel - only one satellite's details are visible at a single time. The panel uses the tabs component (#36) to switch between details, key events, and analysis.

## Panel
- [x] Focus Button: Clicking on this will zoom the visualizer to that satellite
  - [x] If no data is available for the satellite at the current timestamp, disable this button
- [x] Pin button: Clicking on this will remove the object from the focus list
- [x] Close Button
- [x] Object Name

## Tabs
- [x] Details (#45)
- [x] Key Events (#46)
- [x] Analysis (#47)

jnschrag:
Close #44 
Also resolves some items in #126 

This PR does a few related tasks:
1. In a situation where a user is viewing the details pane of a satellite that does not have orbital information for the given timeframe, it hides the orbital information and instead shows the message "This object is not yet in orbit." that is used elsewhere on the site.
2. In the above situation, the button to "focus" on the satellite is disabled. The button to "pin" the satellite is still active, because we do want someone to be able to add the satellite to their focus list.
  *  There is a weird side effect that can happen here: "pinning" a satellite adds it to the focus list, which can cause the visualizer to update to only show "focused" satellites. In this case, the visual is empty because there's no orbital information available for the pinned satellite (if it's the only one pinned). The disclaimer is included in both the details pane and the focus list though, so I think this is an acceptable trade-off. Just wanted to flag as to why this button was not also disabled.
3. This also fixes a bug where if the satellite didn't have any orbital information the whole site would crash because there weren't proper conditionals in place to handle that use case.

## To Test
1. When all the satellite's have loaded, find a satellite that is launched after 2011.
2. With the details pane of that satellite open, change the time of the timeline to sometimes in 2010.
3. The visual will update with new orbital data and you should see the above changes.

### DIFF
--- a/components/visualizer/details-panel/Details.vue
+++ b/components/visualizer/details-panel/Details.vue
@@ -51,9 +51,9 @@
         </dd>
       </div>
     </dl>
+    <hr />
+    <h3>Orbit</h3>
     <div v-if="hasOrbit">
-      <hr />
-      <h3>Orbit</h3>
       <p class="details-panel__small-desc">Last updated on {{ orbitSource }}</p>
       <dl class="details__orbit">
         <div v-for="item in info.orbit" :key="item.value">
@@ -73,6 +73,9 @@
         </div>
       </dl>
     </div>
+    <p v-else class="details-panel__small-desc">
+      This object is not yet in orbit.
+    </p>
     <!-- <hr />
     <h3>ITU Filings</h3> -->
     <template v-if="satellite.acf.comments">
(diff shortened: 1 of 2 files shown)


HP4k1h5 (approved):
looks good!

PR #183, opened 2021-07-15
Refactor/magic improvements

HP4k1h5:


assets/css/components/modal.scss
```diff
@@ -2,6 +2,7 @@
   display: flex;
   flex-direction: column;
   overflow-x: auto;
```
jnschrag, line 4: This can be simplified to `overflow: auto`.

components/visualizer/MagicChart.vue
```diff
@@ -1,5 +1,8 @@
 <template>
-  <Modal @close="updateMagicChart({ showMagicChart: false })">
+  <Modal
+    class="magic-chart modal"
```
jnschrag, line 3: We need to modify this a bit. The `.modal` class is being applied to two elements, the containing div and the `role="dialog"` div, which opens this up to conflict with each other. I think we can probably remove `modal` from the `<Modal>` tag, but take a look and let me know what you think.

assets/css/components/magic-chart.scss
```diff
@@ -1,8 +1,8 @@
 @import '~/assets/css/components/highcharts-theme';
 
-.magic-chart {
-  min-width: 45vw;
-  min-height: 45vh;
+.magic-chart .modal {
+  width: 45vw;
```
jnschrag, line 4: Can we try this instead?
```
width: 100%;
max-width: 680px;
max-height: 85vh;
margin: rem(16);
```

Using `vw` for the width units here causes it to shrink a little oddly as the screen gets smaller.

jnschrag (changes requested):

assets/css/components/magic-chart.scss
HP4k1h5: had a tough time getting the right behavior with these. setting a width was shifting the modal all the way left in the screen

jnschrag (approved):

PR #187, opened 2021-07-20
Feat/itu filings

HP4k1h5:
mostly there. below is a screen of the data table. ive applied some css styles, but wanted some guidance on how to handle long "bands" lists, they can be bigger than this, e.g. 
```
C, K, Ka, Ku, L, S, ULF, V, VHF, W, X
```

possible partial solutions include
- reducing font size/weight/width
- allowing horizontal scroll in this table field
- hovering for full info
- stacking these vertically like what we have now

Screen Shot 2021-07-20 at 4 38 27 PM


assets/css/components/itu-block.scss
```diff
@@ -0,0 +1,41 @@
+.itu {
+  &-list {
+    margin: 0;
+    padding: 0;
+  }
+
+  &__basic {
+    .itu-list & {
```
jnschrag, line 8: I don't think the `.itu-list` class is used anywhere in the code base, is it? If so, it can probably be removed.

jnschrag (changes requested):
Overall this looks really good. I've got a couple of styling tweaks, but functionally I think this is great.

On your question about how to handle the overflow on the table, I think that's probably a question you should ask Tucker in the #aerospace channel in Slack and see what she says.

- [x] Can we make this section heading, "Nearby ITU Filings"?
- [x] Between the section heading and the table, can we add this microcopy (use the same styles as the "Updated from..." text in the Orbit section). `This list includes ITU filings near [%Longitude].` Where `[%Longitude]` is the longitude we're using to populate the table?
- [x] The column headings for the table should have these styles applied:
```
color: $color-white-160;
@extend %font-ui-text-sm-oxygen;
font-weight: bold;
text-transform: uppercase;
```
- [ ] There should be 16px of padding on the right & left side of the table rows. Right now the longitude number runs up right against the edge (might be impacted based on what Tucker has to say about the size of the table)
- [ ] Can the link icon be centered under the "Link" table heading? Can it also change to `$color-white-100` on hover?

HP4k1h5:
The `external-link` icon has color applied in the svg and this is preventing color changes on hover. we'll have to re-import this svg without coloring applied

assets/css/components/itu-block.scss
```diff
@@ -0,0 +1,47 @@
+.itu {
+  &__basic {
+    @extend %font-ui-text-sm-oxygen;
+    grid-template-columns: auto;
```
jnschrag, line 4: The `grid-template-columns`, `@include shadow` and hover states can all be removed from this selector. They aren't used here.

assets/css/components/itu-block.scss
```diff
@@ -0,0 +1,47 @@
...
+
+    &:hover {
+      background: $color-slate-300;
+      @include shadow(sm);
+    }
+  }
+
+  &__longitude {
+    @extend %font-ui-text-lg-oxygen;
+    margin-left: 16px;
```
jnschrag, line 18: Can you switch this to `rem(16)` for consistency? And can we apply it on the `<th>` as well?

assets/css/components/itu-block.scss
```diff
@@ -0,0 +1,47 @@
...
+    @extend %font-ui-text-base-oxygen;
+    color: $color-white-180;
+  }
+
+  &__bands {
+    @extend %font-ui-text-sm-oxygen;
+    color: $color-white-180;
+  }
+
+  &__link {
```
jnschrag, line 32: The last column in the table should have the same amount of padding on the outside as the first column does.

assets/css/components/itu-block.scss
```diff
@@ -0,0 +1,47 @@
...
+    justify-content: center;
+    align-content: center;
+
+    svg {
+      position: relative;
+      top: 3px;
+    }
+
+    // TODO: won't work until svg is re-imported
+    &svg:hover {
```
jnschrag, line 43: We can update this to so it's applied when we fix the SVG.

```
&__link {
  a {
    color: $color-orangedull-100;

    &:hover {
      color: $color-white-180;
    }
}
```

components/visualizer/details-panel/Details.vue
```diff
@@ -76,8 +76,50 @@
     <p v-else class="details-panel__small-desc">
       This object is not yet in orbit.
     </p>
-    <!-- <hr />
-    <h3>ITU Filings</h3> -->
+    <hr />
+    <h3>Nearby ITU Filings</h3>
+    <span class="details-panel__small-desc"
```
jnschrag, line 81: This should be a `<p>` tag. And can we add a `.` at the end of the sentence? Thanks!

components/visualizer/details-panel/Details.vue
```diff
@@ -76,8 +76,50 @@
...
+            :href="props.row.link"
+            target="_blank"
+            aria-label="View ITU Filing"
+          >
+            <Icon id="link" name="external-link" focusable="false" />
+          </a>
+        </span>
+      </template>
+    </vue-good-table>
+    <div v-else class="details-panel__comments">
```
jnschrag, line 120: This should also probably be a `<p>` tag, and should use a different class, since `.details-panel__comments` is specifically for the comments section. We don't want to cross-contaminate styles accidentally.

jnschrag (changes requested):
Thanks for fixing this up. Just a few more changes. I pinged Tucker in Slack so hopefully she'll look at this next week.

**Edit:** How do you feel about setting a max height of the ITU filings table, say ~400px, and having there be an internal scroll? Or is that too much scroll within a scrollable area?

jnschrag (approved):
This looks good to me. I'm going to leave it unmerged for now until @tuckabelle has a chance to look at it next week. Nice job, @HP4k1h5!

PR #188, opened 2021-07-20
Track entity and highlight orbits when query string link is navigated to

samadd:
This required slightly more tweaking that first anticipated, but I'm fairly confident that it now handles the various scenarios (ie user clicks nuxt link in app, or navigates direct to link with query string, or clicks link that changes satellite IDs but not date which doesn't trigger reload of data, or clicks link that changes date and does trigger reload.

There may be some residual quirks waiting to be discovered. I had to change a place in PanelLeft.vue where the reset button was using history.replacestate, because although the query string params would get cleared, the last params would still be there as far as the app was concerned - not sure if that's a Nuxt bug or because of the history.replacestate within the client-side conditional. Anyway, I changed it to use $router, which seemed to clear the query string param state everywhere.


jnschrag (changes requested):
This is looking really good, @samadd. One of behavior that is still off:

If you click on one of the close approaches so it zooms in to show just those two satellites, and then click the "reset" button. It seems like those two satellites are still "focused". The details pane remains open, and the Cesium camera remains locked in place as opposed to resetting like it does now. The rest of it works great.

jnschrag (approved):
Looks good to me, thanks Sam!

PR #189, opened 2021-07-22
Feat/magic chart print styles

jnschrag:
Works on the print & export styles for the Magic Chart. Still need to add the Satellite Dashboard logo and tweak a few things, but this is a good start.

### DIFF
--- a/components/visualizer/MagicChart.vue
+++ b/components/visualizer/MagicChart.vue
@@ -84,7 +84,35 @@ export default {
         },
         exporting: {
           enabled: false,
-          allowHTML: true
+          allowHTML: true,
+          chartOptions: {
+            chart: {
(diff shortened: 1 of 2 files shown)


components/visualizer/MagicChart.vue
```diff
@@ -77,15 +77,53 @@ export default {
...
+              enabled: false
+            }
+          }
+        },
         credits: { enabled: false },
         legend: {
-          enabled: true
+          enabled: true,
+          labelFormatter: function() {
+            console.log(this)
```
HP4k1h5, line 124: rm `console`

HP4k1h5 (changes requested):
really great work! looks good across screen sizes etc.

PR #191, opened 2021-08-03
On date change, grab snapshot of showing orbital paths

samadd:
We listen for the input date change, and grab a snapshot of all orbital paths that are set to show, and then use that when the data comes in in the same way that orbital paths are shown if the query string is set to contain satellite ids.

### DIFF
--- a/components/visualizer/CesiumViewer.vue
+++ b/components/visualizer/CesiumViewer.vue
@@ -144,9 +144,9 @@ export default {
       },
       focussedSatellites: new Set([]),
       highlightedEntities: new Set([]),
-      activeOrbitPaths: [],
       entityLabel: null,
-      labelContainer: null
+      labelContainer: null,
+      lastSelectedEntity: null
     }
   },
   computed: {
@@ -165,7 +165,6 @@ export default {
   watch: {
     satelliteOrbits: 'processNewData',
     visibleSatellites: 'toggleObjectVisibility',
-    selectedDate: 'mapHighlightedOrbits',
     '$route.query.time': 'handleTimeQuery',
     '$route.query.satids': 'handleSatIdChange'
   },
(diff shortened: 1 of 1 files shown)


HP4k1h5:
Thanks, Sam for looking into this. It works well when one satellite is clicked. im able to go back/forth through time and the satellite orbit remains highlighted. However, when more than one satellite orbit is shown, going back/forth in time causes essentially everything to disappear. have you seen this behavior?

looking over the code, and doing some debugging, i haven't found the source of this behavior

samadd:
@HP4k1h5 - thanks. This is very odd behaviour indeed. I've been looking at it this evening, but yet to figure out why having more than one highlighted entity would break things! Even stranger, if I go from 6th August to 1st with 2+ entities selected, then nothing else shows except for when I scroll the timeline back to the 6th, and magically they show up again...

I've a few more ideas to explore on the basis of what I've seen tonight, so should have it fixed soon!

HP4k1h5:
@samadd very grateful for your work on this. i also noticed some of the quirks you're describing. i'll look into it further in a bit also.

samadd:
The strangeness continues... on this branch, what seems to be happening is that when changing date, I see only a handful of satellites on the newly selected date, but if I move the date slider back to the previously selected date, then all the satellites re-appear. So today's date selected, select a few days back, then move back towards today's date. This seems to be independent of whether I've got anything selected or not, but I think is the cause of the behaviour observed. The main branch is working fine, so I'm going to start from there and see how it functions with the latest changes in main.

samadd:
I think I've *finally* fixed(?) it... It seemed to be something to do with references to entities still remaining in the component that had been removed from the entity collection. So highlightedEntities, for example, might refer to entites that should be gone after viewer.entities.removeAll() gets called, and somehow when there are more than one(?) Cesium silently breaks, leaving no trace of an error.
So to fix this I've resorted to clobbering everything after first grabbing a list of any entities with showing paths (which I seemingly have to use the internal _value property to get the boolean inside the Cesium wrapper class).

HP4k1h5:
@samadd Thank you for sorting this. I've noticed a number of buggy interactions have been cleaned up by these changes. Really nice work!

HP4k1h5 (approved):
🥳

PR #204, opened 2021-08-06
edit: fix panel content transition; updates open

### LINKED ISSUE #202: On load the "Latest Updates" should be open

### LINKED ISSUE #203: Improve the panel close animation, please. Right now the body of the panel disappears, but the tabs slide over. If both could slide, that would be great!

### LINKED ISSUE #206: Update the pin icon (the one that looks like it is only outlines) so it isn't just a stroke. I made a new one on the icons page that you can use.

### LINKED ISSUE #207: There are a lot of places where blue outlines (focus states) appear when clicked..
. I'm not sure we need to have that everywhere. For example.
- when someone clicks the panel background, the whole panel is outlined
- Inside of Details window

HP4k1h5:
Close #203
Close #202
Close #206 
Close #207


static/symbols.svg
```diff
@@ -108,9 +108,9 @@
 <title>pen</title>
 <path opacity="0.827172" d="M15.413 1.522c0.192-0.187 0.455-0.286 0.708-0.286 0.263 0 0.516 0.089 0.718 0.286l2.366 2.304c0.189 0.184 0.296 0.434 0.296 0.694s-0.106 0.51-0.296 0.694l-1.85 1.802-3.792-3.693 1.85-1.802zM0.5 15.928l11.529-11.223 3.909 3.805-11.529 11.223h-3.909v-3.805z"></path>
 </symbol>
-<symbol id="icon-pin" viewBox="0 0 20 20">
+<symbol id="icon-pin" viewBox="0 0 26 25">
```
jnschrag, line 111: We want to replace the `icon-unpin` instead (the one that is not filled in). The `icon-pin` is the filled in version.

jnschrag (changes requested):
Left a few comments on specific items!

jnschrag (approved):
The pin change looks good. Do you want to remove the reference to the focus states and one of us can look at that as a part of its own issue? I'm happy to take that on and go through and individually remove them. This is good to go and merge in, but I'll wait until the reference to that issue is removed so we don't accidentally close it/forget about it.

HP4k1h5:
> Do you want to remove the reference to the focus states and one of us can look at that as a part of its own issue?

@jnschrag there are still [two removed outlines](assets/css/components/panel--left.scss) in the `panel-left` scss. 

do you want me to restore these as well?

jnschrag:
> > Do you want to remove the reference to the focus states and one of us can look at that as a part of its own issue?
> 
> @jnschrag there are still [two removed outlines](assets/css/components/panel--left.scss) in the `panel-left` scss.
> 
> do you want me to restore these as well?

Oops, I missed that. Nope, this looks good. There might be others, and we'll have to catch those as we go! Gonna accept and merge this in.

PR #222, opened 2021-08-10
Remove unused ACF fields

### LINKED ISSUE #121: Override Satellite Information
Several of the fields for the satellite information have the ability to be manually overridden in WordPress. Those are currently pulled in via `getSatellites` in `store/satellites.js`, but the values are never used. If those fields have a value, they should override their corresponding field.

## Example
- Default data: `Name: Luch`
- Override data: `acf.name: Luch 17`
- Desired data: `Name: Luch 17`

jnschrag:
Closes #121 

The front end companion to [Override AG Data](https://github.com/CSIS-iLab/satellite-dashboard-backend/pull/34). This updates the satellites object in the store to only keep the ACF fields that we explicitly need - the catalog_id and comments fields.

There are a few remaining references to `acf.catalog_id`. These occur in reference to a post's related satellites ACF, so couldn't be totally removed as they're used to match up the related satellite with its corresponding entry in the satellites store.

### DIFF
--- a/components/visualizer/details-panel/Details.vue
+++ b/components/visualizer/details-panel/Details.vue
@@ -78,11 +78,11 @@
     </p>
     <!-- <hr />
     <h3>ITU Filings</h3> -->
-    <template v-if="satellite.acf.comments">
+    <template v-if="satellite.comments">
       <hr />
       <h3>Comments</h3>
       <ul
-        v-for="comment in satellite.acf.comments"
+        v-for="comment in satellite.comments"
         :key="comment.date"
         class="details-panel__comments"
         role="list"
--- a/store/satellites.js
+++ b/store/satellites.js
@@ -183,8 +183,6 @@ export const actions = {
 
       /**
        * Todo:
-       * Show manual overrides in ACF fields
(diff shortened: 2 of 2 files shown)


HP4k1h5:
looks good. is the comment on line 186 of [store/satellites.js](https://github.com/CSIS-iLab/satellite-dashboard/pull/222/files#diff-733a7c85a541750ce9c7c65c342e2ba1e35d7d7a0fc5520523743a02747f43b0) deprecated by these changes?
```
       * Show manual overrides in ACF fields
```

jnschrag:
> looks good. is the comment on line 186 of [store/satellites.js](https://github.com/CSIS-iLab/satellite-dashboard/pull/222/files#diff-733a7c85a541750ce9c7c65c342e2ba1e35d7d7a0fc5520523743a02747f43b0) deprecated by these changes?
> 
> ```
>        * Show manual overrides in ACF fields
> ```

Yep! Just pushed a change for the alternate name field and removed that Todo comment. Good catch.

HP4k1h5 (approved):
awesome!

PR #226, opened 2021-08-11
Fix/timeline UI

### LINKED ISSUE #200: The "scale" dropdown should have the hand cursor (like the "speed" dropdown)

### LINKED ISSUE #208: The tab for a panel that is already open does not need a hover state

### LINKED ISSUE #210: The user doesn't need to be able to click the "Edit" button if they don't have anything pinned

### LINKED ISSUE #211: Make the whole object block clickable (opening the details window)

### LINKED ISSUE #228: The timeline play button doesn't seem to work the first time I press it. It worked once I updated the Speed and pressed play again.

HP4k1h5:
Close #200 
Close #208
Close #210  
Close #211 
Close #228

### DIFF
--- a/assets/css/components/panel--left.scss
+++ b/assets/css/components/panel--left.scss
@@ -75,7 +75,7 @@
         drop-shadow(0 0 1px rgba(0, 0, 0, 0.44));
       transition: background 0.3s ease-in-out;
 
-      &:hover {
+      &[aria-selected='false']:hover {
         --tab-bg: #{$color-slate-400} !important;
         --tab-cut-out: invert(9%) sepia(9%) saturate(1146%) hue-rotate(169deg)
           brightness(92%) contrast(90%) !important;
--- a/assets/css/components/satellite-block.scss
+++ b/assets/css/components/satellite-block.scss
@@ -17,6 +17,7 @@
       &:hover {
         background: $color-slate-300;
         @include shadow(sm);
+        cursor: pointer;
       }
     }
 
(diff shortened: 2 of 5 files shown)


tuckabelle:
Well, there's a hover state for the blocks, so we could just attach the outline for the focus state, I think.

As for **Option 1**, would it work better if we:
1. Removed the "Info" icon
2. Make the Name the button's phrasing content
3. Made the dot and the Norad ID pseudo elements
?

jnschrag:
> Well, there's a hover state for the blocks, so we could just attach the outline for the focus state, I think.
No, we can't. Because it's not a link or a button item, it's not natively tabbable. With the changes Bob made, it's clickable, but there's no focus state because it isn't a tabbable element. Making it one puts us into Option 2 territory.


> As for **Option 1**, would it work better if we:
> 
> 1. Removed the "Info" icon
> 2. Make the Name the button's phrasing content
> 3. Made the dot and the Norad ID pseudo elements
>    ?

Not at this point - we'd have to rebuild this whole component from scratch to do this.

jnschrag:
@HP4k1h5 I had a thought about this morning and I think we can split the difference here. We can leave the event on the list item like you've changed in this PR, but also leave the information icon a button with the event attached. That way folks who are using their keyboard or a screen reader can still access the event, but the overall card is still clickable.

HP4k1h5:
@jnschrag i think this latest push addresses this issue. `.stop` prevents propagation of the click to the parent <li> when clicking on the info button

jnschrag:
> @jnschrag i think this latest push addresses this issue. `.stop` prevents propagation of the click to the parent
> 
> * when clicking on the info button

The icon still needs to be a button in order to address the accessibility issue, though. You lose the ability to tab to it if it's a `<div>`.

jnschrag (approved):

PR #231, opened 2021-08-12
Added satellite block variation

### LINKED ISSUE #219: Align the sort button to the right

jnschrag:
Closes #219 

Makes a new variation on the `.sat__basic--status` block with a simpler grid structure for just the status icon and the satellite name. I applied those in a couple of other places where those were the only items listed.

I also explicitly set the max width on the filter results version, which fixes the alignment issues as far as I can tell. There's likely to still be some screen sizes where things get a little bit unaligned because of how Vue Table works and calculates the widths of the table columns, but this is the best we can do with the existing structure.

### DIFF
--- a/assets/css/components/satellite-block.scss
+++ b/assets/css/components/satellite-block.scss
@@ -26,6 +26,15 @@
       column-gap: var(--marker-gap, rem(16));
       row-gap: rem(2);
 
+      &-simple {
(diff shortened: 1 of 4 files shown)


HP4k1h5 (approved):
looks good to me. @tuckabelle does this look right to you

PR #234, opened 2021-08-20
Tooltip should follow mouse position

### LINKED ISSUE #230: The tooltip that accompanies the "info" panel should appear near the user's mouse

jnschrag:
Closes #230 

I set the [`follow-cursor` attribute](https://kabbouchi.github.io/vue-tippy/4.0/demo.html#follow-cursor) for the Tippy tooltips so it should more closely follow the mouse.

### DIFF
--- a/components/global/GlossaryTooltip.vue
+++ b/components/global/GlossaryTooltip.vue
@@ -8,6 +8,7 @@
         interactive-border="4"
         distance="12"
         arrow
+        follow-cursor="true"
       >
         <template v-slot:trigger>
           <button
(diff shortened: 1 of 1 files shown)


components/global/GlossaryTooltip.vue
```diff
@@ -8,6 +8,7 @@
         interactive-border="4"
         distance="12"
         arrow
+        follow-cursor="true"
```
HP4k1h5, line 11:
```suggestion
        :follow-cursor="true"
```

PR #245, opened 2021-08-27
edit: use min_distance and fmt in place

HP4k1h5:
there is no min_distance_km in the db

closes https://www.notion.so/csisilab/Kaitlyn-b49b91c56d8b49889e12d30630cc4832

### DIFF
--- a/pages/key-events.vue
+++ b/pages/key-events.vue
@@ -148,8 +148,8 @@
                 </div>
               </div>
             </template>
-            <template v-else-if="props.column.field === 'min_distance_km'">
-              {{ formatNumber(props.row.min_distance_km) }}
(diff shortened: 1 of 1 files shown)


HP4k1h5:
> This works, but my one thought would be is if this is a change we should actually make on the server side. We could check to see if [the sort param](https://github.com/CSIS-iLab/satellite-dashboard-backend/blob/main/wp-content/plugins/satellite-dashboard/api/close-approaches.php#L93-L100) is by `min_distance_km` and update it there, so that way the km formatting and all of that stays on the server side.
> 
> Up to you. If you're happy with doing it this way on the FE, then feel free to merge this in!

id rather leave the backend as is for now.

PR #246, opened 2021-08-30
edit: adjust post sat table, v align status dot

HP4k1h5:
closes https://www.notion.so/csisilab/Thomas-G-Roberts-8b80eb8602084043912d7f33ec341b62


jnschrag (approved):
Closes #247 

@HP4k1h5 I modified this so it vertically aligns the status symbol consistently everywhere this block is used. It also takes care of vertical aligning the country names & any "actions" in that row.

PR #265, opened 2022-01-28
Add social images from analysis posts

jnschrag:
Give this a try. It pulls in the featured image from a post on WP. I also moved the site wide social image to the static directory. I couldn't test it fully because of the local URL, but you can test a URL on the [Twitter card validator](https://cards-dev.twitter.com/validator) to confirm that this works.


urchykli (commented):
Thanks so much for this, Jacque!

urchykli (approved):

PR #279, opened 2023-08-22
Update dependencies to remove Fiber and old sass deps

samadd:
Updates some dependencies that were forcing the project to be run on Node 14.21.3. Now runs successfully on Node 18 after removing Fibers (which is no longer supported) and updating to latest Nuxt 2 version (not Nuxt 3).


jdromero88:
@samadd thanks for working on this! Before merging this into main, we would like you to check out a weird bug. When you leave the visualizer to go to analysis for example and come back to the visualizer, will sometimes work, other times won't show any satellites, not even an error on the console, other times won't load any satellites, but it shows this error:
image
I didn't dig into the error, but please let us know your findings.

Also, would be great if you can work on the next error. When visiting the key events page throws these errors:
image

Here is the link to your deployed [branch](https://feature-node-18-compatibility--satdash.netlify.app/). Please let us know if you need anything.

jdromero88 (changes requested):
See this [comment](https://github.com/CSIS-iLab/satellite-dashboard/pull/279#issuecomment-1734296581)

samadd:
Hi @jdromero88 - I've pushed a couple of commits that should deal with the 2nd error(s). The first screenshot you provided, however, seems to show what looks like a temporary issue with the CDN provider. I haven't been able to recreate it, at least. A CORS error would be on the unpkg.com side, and most likely a side effect of another error (see the 524 timeouts) on their side, in that an error response without the correct CORS headers would also result in a CORS error.

jdromero88:
Thank you @samadd! I will review this ASAP

jdromero88 (approved):

